PR #2, opened 2017-03-21
Add user agent

wgreenberg:
Some websites refuse to serve content to requests without one, and also
it's just good manners.


sgnn7:
CR pass

PR #5, opened 2017-03-29
Added icon for the previewer

sgnn7:
Blank icon was pretty unprofessional so we now have one for the
previewer.


nedrichards:
lgtm

PR #7, opened 2017-03-30
Libingester: Added a test case for wowshack and cleaned up the code a bit

sgnn7:
https://phabricator.endlessm.com/T16359

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -174,11 +174,20 @@ function is_subset(a, b) {
 }
 
 class Hatch {
-    constructor() {
+    constructor(name) {
         this._promises = [];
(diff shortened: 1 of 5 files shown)


lib/index.js
```diff
@@ -18,6 +18,9 @@ function new_asset_id() {
 class ValidationError extends Error {
 }
 
+class InvalidArgumentError extends Error {
```
wgreenberg: this isn't used anywhere

sgnn7: Ugh - originally I made no name specified to throw this error and later swapped to no error but didn't remove this.

PR #12, opened 2017-05-03
Libingester: Add support for canonical_uris for datauri assets

sgnn7:
https://phabricator.endlessm.com/T16375


test/lib/util.test.js
```diff
@@ -47,24 +47,55 @@ describe('download_image', function() {
...
-        expect(asset.asset_id).is.not.null;
-        expect(asset.asset_id).is.not.undefined;
-        expect(asset._image_data).to.deep.equal(image);
-        expect(asset._content_type).to.equal('image/png');
+            const asset = util.download_img(imageTag, 'a://b.com/c d.html');
+            expect(asset.asset_id).is.not.null;
+            expect(asset.asset_id).is.not.undefined;
+            expect(asset._image_data).to.deep.equal(image);
+            expect(asset._content_type).to.equal('image/png');
+            expect(asset._canonical_uri).to.equal("data:image/png;" +
```
wgreenberg: generally I'm against testing private variables like this -- what about turning `_to_metadata()` public and testing these fields in its output?

sgnn7: Very good point - will fix.

wgreenberg:
LGTM!

PR #14, opened 2017-05-04
Word of day

syndacks:


examples/word_of_day.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+const libingester = require('libingester');
+const mustache = require('mustache');
+const rp = require('request-promise');
+const template = require('./word_of_day_template');
```
sgnn7: Generally, you want the template body to be the requirement, not the code to load it.

i.e. You will probably want to put only the template into a separate file and load it inline in this one.

examples/word_of_day.js
```diff
@@ -0,0 +1,80 @@
...
+        // use url library to pull out query string date object
+        const parts = url.parse(uri, true);
+        const modified_date = new Date(Date.parse(parts.query.d));
+
+        asset.set_last_modified_date(modified_date);
+        asset.set_section("word_of_day");
+
+        //Set title section
+        const date = parts.query.d;
+        const title = $profile('h1').first().text() + ": " + date;
```
sgnn7: Try to keep items grouped a bit better (vars on top and logic in the middle). title/date here should probably be near the top of this method since they get set once and used at the end.

examples/word_of_day.js
```diff
@@ -0,0 +1,80 @@
...
+
+        // Pluck wordOfDay
+        const wordOfDay = $profile('table.widget', 'div#colleft').first().find('h3').children().text();
+        asset.set_title(wordOfDay);
+
+        // Pluck wordOfDayDef
+        const wordOfDayMixedString = $profile('td:contains("Definition:")').next().text();
+        const wordOfDayDef = wordOfDayMixedString.split(" ").slice(1).join(" ");
```
sgnn7: This seems too messy of logic. Try to slice the string from first space onward with `indexOf()` as the begging.

examples/word_of_day.js
```diff
@@ -0,0 +1,80 @@
...
+
+        asset.set_synopsis(wordOfDayDef);
+
+        // Pluck wordOfDayType
+        const pos = wordOfDayMixedString.split(" ")[0];
```
sgnn7: Needs a better variable name

examples/word_of_day.js
```diff
@@ -0,0 +1,80 @@
...
+        const wordOfDayType = pos.substr(1, pos.length-2);
+
+        asset.set_license(wordOfDayType);
```
sgnn7: Group these setters together

examples/word_of_day.js
```diff
@@ -0,0 +1,80 @@
...
+    });
+}
+
+function main() {
+    const hatch = new libingester.Hatch();
+    // make request to the index (home) page
+    libingester.util.fetch_html(home_page).then(($pages) => {
+        // retrieve all the URLs for individual pages
+        const articles_links = $pages('#Calendar div:nth-child(-2n+2) a').map(function() {
+        // const articles_links = $pages('#Calendar div:nth-child(-n+15) a').map(function() {
```
sgnn7: Stray comment?

lib/index.js
```diff
@@ -151,6 +152,7 @@ class NewsArticle extends BaseAsset {
     to_metadata() {
         const metadata = super.to_metadata();
         metadata['document'] = this._document.toString();
+        metadata['synopsis'] = this._synopsis.toString();
```
sgnn7: What if a base asset doesn't have a synopsis set?

examples/word_of_day.js
wgreenberg: @sgnn7 we use this convention everywhere else, IMO it's totally fine since it's just a shortcut to `fs.readFileSync`

examples/word_of_day.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+const libingester = require('libingester');
+const mustache = require('mustache');
+const rp = require('request-promise');
+const template = require('./word_of_day_template');
+const url = require('url');
+
+const home_page = 'http://www.thefreedictionary.com/_/archive.htm'; // Home section
+
```
wgreenberg: nit: no need for double newlines

examples/word_of_day.js
```diff
@@ -0,0 +1,80 @@
...
+        hatch.save_asset(asset);
+    });
+}
+
+function main() {
+    const hatch = new libingester.Hatch();
+    // make request to the index (home) page
+    libingester.util.fetch_html(home_page).then(($pages) => {
+        // retrieve all the URLs for individual pages
+        const articles_links = $pages('#Calendar div:nth-child(-2n+2) a').map(function() {
```
wgreenberg: I think the `nth-child(-2n+2)` could use some explaining in the comment

examples/word_of_day.js
sgnn7: Eh I'm not really a fan of ithat approach since then you have to define your template in code rather than a separate file but it's not a big deal.

examples/word_of_day.js
syndacks: unless i'm misunderstanding, you can't pass `indexOf()` as a parameter to slice such as `str.slice(indexOf(" "))`

you could do something like:
`const indexOfBlah = wordOfDayPre.indexOf(" ");`
`const wordOfDayDef = wordOfDayPre.slice(indexOfBlah);`

but this seems a bit messy too eh? @sgnn7

syndacks:
thanks for the comments guys.
i'm going to come back to the other ones later.

examples/word_of_day.js
sgnn7: Inline it: `const wordOfDayDef = wordOfDayMixedString.slice(wordOfDayMixedString.indexOf(" "));`

syndacks:
Got it. Small things like this I'm still learning!  Thanks.

...................................................................

Dacks Milliken  | +1. <http://goog_1649640352>650.822.7329
<%2B1.650.822.7329>  |  Endless <http://endlessm.com/>

lib/index.js
syndacks: @sgnn7 I think I know what you're asking...which is, if there is no 'synopsis' value set, `this._synopsis.toString()` will return undefined and throw an error. I experienced something like this when I was working with the `document` field (I didn't pass it a value) and I got some unreturned promise errors.

Assuming this is what your'e getting at, would some conditional logic in the ingester to set synopsis to empty string (if no synopsis, synopsis = "") be a workaround? Furthermore, would similar types of logic needed to be placed in the code for the various other fields, like `document`? But because I don't see any such logic elsewhere, now I'm wondering if this line of thinking is accurate.

sgnn7:
@syndacks Is this ready for a re-review?

syndacks:
@sgnn7 did you see my comment above on lib/index.js? if you can give me a bit of direction I can implement the feedback and hopefully get another review. thanks!

lib/index.js
sgnn7: @syndacks Yes, that's the exact problem I'm talking about. Without a document, a news asset is pretty much not working properly to begin with so it's fine if errors happen there but for synopsis, if there isn't one, you can make sure that you don't set one on the metadata synopsis.

PS: I know it might be a bit much to ask right now but there are some tests that you can update to make sure this code works as you expect it: https://github.com/endlessm/libingester/blob/master/test/lib/index.test.js

You can run the tests with: https://github.com/endlessm/libingester/blob/master/run_tests.sh

syndacks:
@sgnn7 : spent some time reading up and learning mocha/chai last night. I updated the test files to reflect the NewsArticle update of 'synopsis' and also added a some logic in index.js to give metadata['synopsis'] an empty string if none is given.

sgnn7:
@syndacks Looks great now!

test/lib/index.test.js
```diff
@@ -44,3 +44,38 @@ describe('ImageAsset', function() {
...
+        asset.set_last_modified_date(new Date(1492545280000));
+        asset.set_document('<h1>Word of the Day</h1>');
+        asset.set_section("word_of_day");
+        asset.set_synopsis('a long time ago...');
+
+        const metadata = asset.to_metadata();
+
+        // Remove randomness -- should probably be a mock if I can
+        // figure out how to use it.
+        delete metadata['assetID'];
```
lmyllari: Thumbnail should be handled here too, otherwise the test will not pass. @sgnn7, @wgreenberg: would you prefer the field not be output at all if no thumbnail is set - like in the change above for synopsis?

PR #15, opened 2017-05-04
T16849: add synopsis and thumbnail field setters

lmyllari:


test/lib/index.test.js
```diff
@@ -27,6 +28,7 @@ describe('ImageAsset', function() {
             "matchingLinks": [ 'https://www.example.com/' ],
 
             "title": 'Test Asset',
+            "synopsis": 'Test Asset synopsis',
```
sgnn7: No test for thumbnail

previewer/metadata_renderer.js
```diff
@@ -12,6 +12,8 @@ const visibleProps = [ 'assetID',
                        'matchingLinks',
 
                        'title',
+                       'synopsis',
+                       'thumbnail',
```
sgnn7: Thumbnails should be good to include in the asset listing on the left pane as well (on the left side of the asset name) when previewing.

examples/olympic.js
```diff
@@ -17,6 +17,10 @@ function ingest_profile(hatch, uri) {
         const modified_date = new Date(Date.parse(modified_str));
         asset.set_last_modified_date(modified_date);
 
+        // Pull out the description.
+        const description = $profile('meta[name="description"]').attr('content');
+        asset.set_synopsis(description);
```
sgnn7: I'd add this to the other two example hatch creation scripts.

lib/index.js
```diff
@@ -74,6 +78,13 @@ class BaseAsset {
             return Promise.all(promises);
         }));
     }
+
+    _get_outgoing_asset_ids() {
+        if (this._thumbnail_asset_id) {
+            return [ this._thumbnail_asset_id ];
+        }
+        return [];
+    }
```
sgnn7: This is a bit tricky here... since all assets types will have this code you're technically implementing `_get_thumbnail_asset_id()` which any other asset that needs to customize will need to override with `super.` invocation.

I think a better way to implement this would be to have `_get_thumbnail_asset_id` (returns thumbnail asset id if there is one), `get_dependent_assets` (that the base has, that is used by APIs, and that has an invocation of `_get_thumbnail_asset_id`), and `_get_dependent_asset_id` (that children can override and add any other non-thumbnail asset id to the list). Let me know on slack if this might not explain things well.

lib/index.js
```diff
@@ -94,7 +105,7 @@ class ImageAsset extends BaseAsset {
 
     to_data() { return this._image_data; }
 
-    _get_outgoing_asset_ids() { return []; }
+    _get_outgoing_asset_ids() { return super._get_outgoing_asset_ids(); }
```
sgnn7: If the method only has the invocation of `super.same_name_func`, I think you don't need it.

lib/index.js
```diff
@@ -114,6 +125,8 @@ class VideoAsset extends BaseAsset {
             title: this._title,
         });
     }
+
+    _get_outgoing_asset_ids() { return super._get_outgoing_asset_ids(); }
```
sgnn7: Ditto on not needing this if you use super

lib/index.js
```diff
@@ -157,9 +170,11 @@ class NewsArticle extends BaseAsset {
...
 
     _get_outgoing_asset_ids() {
-        return this._document.find('[data-soma-job-id]').map(function() {
-            return cheerio(this).attr('data-soma-job-id');
-        }).get();
+        const ids = super._get_outgoing_asset_ids();
+        this._document.find('[data-soma-job-id]').each(function() {
+            ids.push(cheerio(this).attr('data-soma-job-id'));
+        });
+        return ids;
```
sgnn7: See my comment about how to maybe better handle this inheritance

test/lib/index.test.js
sgnn7: Also would be good to add tests for the asset id processing

lib/index.js
lmyllari: Thank you, that sounds like a much better way to do this

previewer/metadata_renderer.js
lmyllari: Good idea and only a few lines of code. I'll add it to the updated PR.

test/lib/index.test.js
lmyllari: Any hints how to go about mocking the asset ID generation?

examples/olympic.js
lmyllari: I think those scripts will need some fixing first. Also, the previewer is missing video support. @wgreenberg: should these be handled as a separate issue?

lmyllari:
@sgnn7 thank you for the review! I've updated the PR, please let me know what you think.

examples/olympic.js
sgnn7: @lmyllari Tests work on current master they just need more coverage for this new functionality:
```
 $ ./run_tests.sh 
<snip>
  10 passing (416ms)
```

test/lib/index.test.js
sgnn7: @lmyllari You don't need mocking for the basics:
For each asset type:
 - New one up, set thumbnail to fake val, and ensure that value returns in get_metadata call
 - Same with synopsis
 - For asset types with extra dependent ids, set a fake doc and verify that get_dependent_assets returns those along with thumbnail

examples/olympic.js
lmyllari: I'll add coverage to the tests, but these example scripts are separate from them?

test/lib/index.test.js
lmyllari: Thank you! I'll update the tests.

examples/olympic.js
sgnn7: Oh - sorry about that I confused this with the other (testing) comment. If we can fix the other examples and add this field, that would be best. We can open stories for them but that would just waste  a ton of time getting it done.

test/lib/index.test.js
```diff
@@ -7,18 +7,23 @@ const libingester = require('../../lib/index');
...
         const metadata = asset.to_metadata();
 
-        // Remove randomness -- should probably be a mock if I can
-        // figure out how to use it.
+        // Check that asset ID and thumbnail asset ID are passed through
+	expect(metadata['assetID']).to.equal(asset.asset_id);
+	expect(metadata['thumbnail']).to.equal(thumbnail_asset.asset_id);
+	// Remove the ID fields before checking the rest
         delete metadata['assetID'];
+        delete metadata['thumbnail'];
```
lmyllari: @sgnn7 do you think this approach is ok?

sgnn7: @lmyllari Indeed - this is good. I don't see any tests for other types of assets but we can deal with that later.

sgnn7:
CR pass though we still need the other two examples fixed at some point.

PR #16, opened 2017-05-10
Fix previewer source/preview switch

lmyllari:
The logic for current and new source display status is inverted, thus
first click on Flip button does nothing. This is fixed by having the
showingSource variable reflect the requested state in flipPage.


sgnn7:
CR pass - good catch!

PR #17, opened 2017-05-12
Remove any 'synopsis' related code from NewsArticle class

syndacks:
There is already a 'set_synopsis' method on NewsArticle's parent, BaseAsset. Therefore, I don't see any need to add any 'synopsis' related methods to NewsArticle, or the need to export any 'synopsis' related metadata in NewsArticle. If I'm overlooking something, please advise. My other line of thinking was to handle a !synopsis issue at the time of ingestion, but I'm going to hold on that for now.


sgnn7:
@syndacks LGTM

PR #18, opened 2017-05-12
16846

wgreenberg:


test/lib/util.test.js
```diff
@@ -104,6 +106,8 @@ describe('download_image', function() {
     });
 });
 
+// FIXME this text intermittently fails because it reaches out to an external
+// server. We should replace it with gzipped content served locally
```
sgnn7: Fair enough - I didn't want to have copyrighted content in our repo and I didn't have a good way to simulate a gzipped server at the time which is why it's a link.

lib/index.js
```diff
@@ -223,6 +223,9 @@ class Hatch {
             for (let outgoing_asset_id of asset.get_dependent_assets())
                 outgoing.add(outgoing_asset_id);
         });
+        this._videos.forEach((video) => {
```
sgnn7: Isn't a video just another "dependent asset" like the list returned by `get_dependent_assets`? In other words, why treat this type differently from the rest?

wgreenberg: yeah this bit's a tad confusing. Videos are dependent assets, and will be present in the `outgoing` set -- however, prior to this commit, they weren't considered part of the hatch's `asset_ids`, so this check failed

sgnn7:
CR pass

PR #21, opened 2017-05-16
Add 'holiday_of_day' ingester and template.

syndacks:
This is the last of 4 ingesters based on FreeDictionary.com content. There are also a few minor tweaks to the 'day_in_history' ingester and template.

### DIFF
--- a/examples/holiday_of_day.js
+++ b/examples/holiday_of_day.js
@@ -3,7 +3,7 @@
 const libingester = require('libingester');
 const mustache = require('mustache');
 const rp = require('request-promise');
-const template = require('./day_in_history_template');
+const template = require('./holiday_of_day_template');
 const url = require('url');
 
 const HOMEPAGE = 'http://www.thefreedictionary.com/_/archive.htm'; // Home section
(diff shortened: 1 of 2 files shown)


examples/day_in_history_template.js
```diff
@@ -11,9 +11,9 @@ const structure_template = (`
 <section class="mainContent">
 
   <h3>On this Day in History</h3>
-    <p id="dayInHistoryHeadFinal">{{ dayInHistoryHeadFinal }}</p>
-    <p id="dayInHistoryBodyFinal">{{ dayInHistoryBodyFinal }}</p>
-    <p id="dayInHistoryYearFinal">{{ dayInHistoryYearFinal }}</p>
+    <p id="dayInHistoryHead">{{ dayInHistoryHead }}</p>
+    <p id="dayInHistoryBody">{{ dayInHistoryBody }}</p>
+    <p id="dayInHistoryYear">{{ dayInHistoryYear }}</p>
```
sgnn7: Generally, messing with other non-related functionality would be in separate commits but not a blocker for now.

examples/holiday_of_day.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
```
sgnn7: Given how similar this one and the other two look, it might be advantageous to combine shared code in a separate module but also not a big deal right now.

sgnn7:
@syndacks LGTM

PR #22, opened 2017-05-16
Output a tarball for a hatch

lmyllari:
Will wrote this, I just rebased and checked the output format. :)


wgreenberg:
I approve of my code!

PR #23, opened 2017-05-17
T16848

wgreenberg:
https://phabricator.endlessm.com/T16848

### DIFF
--- a/lib/soma_dom.js
+++ b/lib/soma_dom.js
@@ -1,0 +1,30 @@
+// This is a collection of data attributes we decorate the DOM with to inform
+// our packer about the semantic use of certain tags. For example, if a link
+// tag which contains an image should be opened in a lightbox on the e-k-l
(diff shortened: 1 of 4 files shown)


lib/soma_dom.js
```diff
@@ -0,0 +1,30 @@
...
+        Link: 'Link',
+        Notes: 'Notes',
+        References: 'References',
+        RelatedLinks: 'RelatedLinks',
+        TableOfContents: 'TableOfContents',
+        Thumbnail: 'Thumbnail',
+        Title: 'Title',
+        Video: 'Video',
+        VideoContainer: 'VideoContainer',
+        VideoLink: 'VideoLink',
```
sgnn7: You could add these programmatically to avoid copy errors but it wouldn't save much space and it's marginal utility so not a blocker.

sgnn7:
CR pass

PR #25, opened 2017-05-19
Make bootstrap css be resolvable if we're imported as a library

sgnn7:
Old code only worked properly when running from the libingester directly
but failed when run as a dependency since paths are different. We now
account for that and try to load either of the stylesheets paths so we
work on both setups.


wgreenberg:
:+1:

PR #26, opened 2017-05-19
Made libingester use timestamped directory names

sgnn7:
Without this, it's really a pain to manage the hatch folders so now wei
add a timestamp instead of a random string to it.


wgreenberg:
lgtm

PR #28, opened 2017-05-22
Added image thumbnails to asset list

sgnn7:
This should make it easier to see the images without even clicking on
them.


milossymphony:
CR Pass

PR #31, opened 2017-05-31
Add blog article format

tchx84:
Introduces a new BlogArticle asset class which
provides a simple API and template to generate
blog articles.

https://phabricator.endlessm.com/T17173

### DIFF
--- a/examples/livingloving.js
+++ b/examples/livingloving.js
@@ -1,0 +1,104 @@
+'use strict';
+
+const libingester = require('libingester');
+const rss2json = require('rss-to-json');
+
+const rss_uri = "http://www.livingloving.net/feed/";
+
+const img_metadata = [
+    'class',
+    'data-jpibfi-indexer',
+    'data-jpibfi-post-excerpt',
+    'data-jpibfi-post-url',
+    'data-jpibfi-post-title',
+    'height',
+    'id',
+    'rscset',
+    'sizes',
+    'src',
+    'width',
+];
+
+const remove_elements = [
+    'iframe',
+    'input',
+    'noscript',
+    'script',
+    '.link_pages',
+    '.jp-relatedposts',
+    '.post-tags',
+    '.sharedaddy',
+    '[id*="more-"]',
+];
+
+function ingest_article(hatch, uri) {
+    return libingester.util.fetch_html(uri).then(($profile) => {
+        let base_uri = libingester.util.get_doc_base_uri($profile, uri);
+
+        let modified_date = $profile('meta[property="article:modified_time"]').attr('content');
+        let article_entry = $profile('.post .post-heading .meta').first();
+        let article_data = $profile(article_entry).text().split(' • ');
+        let author = article_data[0];
+        let date_published = article_data[1];
+        let category = article_data[2];
+        let title = $profile('meta[property="og:title"]').attr('content');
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -170,6 +170,119 @@ class VideoAsset extends BaseAsset {
...
+<header>
+    <div class="categories" id="categories-links"></div>
+    <div class="extra-header-right">
+        <span class="authors">${this._author}</span>
+        &mdash;
+        <span class="date-published">${this._date_published}</span>
+    </div>
+</header>
+<article>
+    <div class="article-title">
```
ptomato: I put my title inside a `<header>` and the article body (including main image) inside a `<section>`, I think that's how the semantic HTML tags are meant to be used.

lib/index.js
```diff
@@ -170,6 +170,119 @@ class VideoAsset extends BaseAsset {
...
+        <h1>${this._title}</h1>
+    <div>
+    <div class="main-image">
```
ptomato: I had it set up so that there is no surrounding element if the main image was missing.

examples/livingloving.js
```diff
@@ -0,0 +1,103 @@
...
+        asset.set_last_modified_date(new Date(Date.parse(modified_date)));
+        asset.set_title(title);
+        asset.set_synopsis(synopsis);
+        asset.set_thumbnail(main_image);
+        asset.set_author(author);
+        asset.set_date_published(date_published);
+        asset.set_license('Proprietary');
+        asset.set_main_image(main_image);
+        asset.set_main_image_caption('Image Caption');
+        asset.set_body(body.html());
```
ptomato: Might be good to follow the pattern of allowing to pass a Cheerio object here, I found that quite useful so that you don't have to re-parse it if you need to manipulate it in the renderer.

tchx84:
@ptomato @sgnn7 updated based on news review and other philip's comments.

lib/index.js
```diff
@@ -170,6 +170,136 @@ class VideoAsset extends BaseAsset {
...
+    set_main_image(value) { this._main_image = value; }
+    set_main_image_caption(value) { this._main_image_caption = value; }
+    set_body(value) { this._body = _ensure_cheerio(value); }
+    set_read_more_text(value) { this._read_more_text = value; }
+    set_tags (value) { this._tags = value; }
+    set_as_static_page() { this._ekn_tags.push(EKN_STATIC_TAG); }
+
+    set_date_published(value) {
+        this._date_published = value;
+        if (!(value instanceof Date))
```
sgnn7: Not a blocker but wrap even single-line conditionals in curlies

lib/index.js
```diff
@@ -170,6 +170,136 @@ class VideoAsset extends BaseAsset {
...
+                }
+
+                if (!$asset.attr('data-soma-job-id')) {
+                    throw new ValidationError("Document has media without associated asset. Please ingest it or drop it.");
+                }
+            });
+        });
+    }
+
+    render() {
```
sgnn7: Same thing I mentioned to @ptomato but this needs to be a real template in a separate file.

sgnn7:
@tchx84 Needs a rebase

sgnn7:
Rebased it myself. CR pass.

PR #32, opened 2017-06-02
NewsArticle: Switch to article format

ptomato:
This introduces the "standard" article format for NewsArticle. This
allows ingesting news articles in a way that can easily be styled
uniformly, with minimal stylesheet changes in between apps.

https://phabricator.endlessm.com/T17174

### DIFF
--- a/examples/ars.js
+++ b/examples/ars.js
@@ -1,0 +1,109 @@
+'use strict';
+
+// Ars Technica, Open Source tag example
+
+const Libingester = require('libingester');
+const thenifyAll = require('thenify-all');
+const RssToJson = thenifyAll(require('rss-to-json'), {}, ['load']);
+const fs = thenifyAll(require('fs'), {}, ['writeFile']);
+
+const FEED_URI = 'http://feeds.arstechnica.com/arstechnica/open-source';
+const COMMENT_NODE = 8;
+
+function remove_intermediate($, selector) {
+    $(selector).each((i, elem) => $(elem).replaceWith($(elem).contents()));
+}
+
+function ingest_article(hatch, {title, description, url, created}) {
+    return Libingester.util.fetch_html(url).then($ => {
+        const BASE_URI = Libingester.util.get_doc_base_uri($, url);
+        let asset = new Libingester.NewsArticle();
+
+        console.log('processing', title);
+        asset.set_title(title);
+        asset.set_synopsis(description);
+        asset.set_date_published(created);
+        asset.set_last_modified_date(new Date(created));
+        asset.set_source('Ars Technica');
+        asset.set_license('Proprietary');
+        asset.set_section('open-source');
+
+        let canonical_uri = $('head link[rel="canonical"]').attr('href');
+        asset.set_canonical_uri(canonical_uri);
+        asset.set_read_more_link(`Read more at <a href="${canonical_uri}">Ars Technica</a>`);
+
+        let thumb_uri = $('meta[property="og:image"]').attr('content');
+        let thumb_asset = Libingester.util.download_image(thumb_uri);
+        hatch.save_asset(thumb_asset);
+        asset.set_thumbnail(thumb_asset);
+
+        // Clean up
+        $('div').filter(function () {
+            return $(this).text().trim() === '';
+        }).remove();
+        $('head, script, .site-header, .ad, #social-left').remove();
+        ['.site-wrapper', '.content-wrapper', '.column-wrapper', '.left-column',
+            '.article-content'].forEach(sel => remove_intermediate($, sel));
+        $('.page-numbers').remove();  // oops, should deal with multiple pages
+        $(`.story-sidebar, .enlarge-link, .post-upperdek, .article-author,
+            #social-footer, #article-footer-wrap, .site-footer, .tools-info,
+            #promoted-comments`).remove();
+        $('*').contents().each(function () {
+            if(this.nodeType === COMMENT_NODE) {
+                $(this).remove();
+            }
+        });
+
+        let authors = $('header [itemprop~="author"] [itemprop~="name"]')
+            .map((i, elem) => $(elem).text())
+            .get();
+        asset.set_authors(authors);
+        $('header').remove();
+
+        let main_candidates = $('figure.intro-image');
+        if (!main_candidates.length)
+            main_candidates = $('figure');
+        if (main_candidates.length) {
+            let main = main_candidates.first();
+            let img = $('img', main);
+            let img_asset = Libingester.util.download_img(img, BASE_URI);
+            hatch.save_asset(img_asset);
+            asset.set_main_image(img_asset, $('figcaption', main));
+            $(main).remove();
+        }
+
+        let first_para = $('section p').first();
+        asset.set_lede(first_para);
+        $(first_para).remove();
+
+        // Clean up classes for readability
+        $('body, article, section, figure, figcaption').removeAttr('class');
+        $('article').removeAttr('itemscope itemtype');
+        $('figure').removeAttr('style');
+
+        // Save assets for any remaining figures
+        $('figure').each(function () {
+            let fig_asset = Libingester.util.download_img($('img', this),
+                BASE_URI);
+            hatch.save_asset(fig_asset);
+        });
+
+        asset.set_body($('section'));
+
+        asset.render();
+        hatch.save_asset(asset);
+    });
+}
+
+function main() {
+    let hatch = new Libingester.Hatch('ars');
+    RssToJson.load(FEED_URI).then(rss =>
+        Promise.all(rss.items.map(entry => ingest_article(hatch, entry))))
+    .then(() => hatch.finish())
+    .catch(err => {
+        console.log('there was an error', err);
+        process.exitCode = 1;
+    });
+}
+
+main();
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,6 +3,7 @@
 const aws = require('aws-sdk');
 const cheerio = require('cheerio');
 const crypto = require('crypto');
+const mustache = require('mustache');
 const path = require('path');
 const verify = require('./verify');
 
@@ -170,6 +171,12 @@ class VideoAsset extends BaseAsset {
 
 exports.VideoAsset = VideoAsset;
 
+function _ensure_cheerio(value) {
+    if (typeof value === 'string')
+        return cheerio.load(value);
+    return value;
+}
+
 class NewsArticle extends BaseAsset {
     constructor() {
         super();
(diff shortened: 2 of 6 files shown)


examples/ars.js
```diff
@@ -0,0 +1,115 @@
+'use strict';
+
+// Ars Technica, Open Source tag example
+
+const Libingester = require('libingester');
+const thenifyAll = require('thenify-all');
+const RssToJson = thenifyAll(require('rss-to-json'), {}, ['load']);
+const fs = thenifyAll(require('fs'), {}, ['writeFile']);
```
sgnn7: I'm really against this pattern as `fs` generally has non-blocking methods but we use it in a lot things in the same manner so not a big deal for now.

lib/index.js
```diff
@@ -170,16 +170,107 @@ class VideoAsset extends BaseAsset {
...
+    }
+
     set_section(value) { this._section = value; }
     set_as_static_page() { this._ekn_tags.push(EKN_STATIC_TAG); }
+    set_source_name(value) { this._source_name = value; }
+    set_lede(value) { this._lede = _ensure_cheerio(value); }
+    set_body(value) { this._body = _ensure_cheerio(value); }
+    set_read_more_link(value) { this._read_more_link = _ensure_cheerio(value); }
+
+    render() {
```
sgnn7: Can we split the template out of the body of the code (this module is going to become untenable if we don't)?

lib/index.js
```diff
@@ -170,16 +170,107 @@ class VideoAsset extends BaseAsset {
...
+    }
+
+    set_main_image(asset, caption) {
+        this._main_image = asset;
+        this._main_caption = _ensure_cheerio(caption);
+    }
+
     set_section(value) { this._section = value; }
     set_as_static_page() { this._ekn_tags.push(EKN_STATIC_TAG); }
+    set_source_name(value) { this._source_name = value; }
```
sgnn7: Any specific reason to use `set_source_name` vs `set_source`?

lib/index.js
```diff
@@ -170,16 +170,107 @@ class VideoAsset extends BaseAsset {
...
+    set_authors(value) {
+        if (value instanceof Array)
+            this._authors = value;
+        else
+            this._authors = [value];
+    }
+
+    set_main_image(asset, caption) {
+        this._main_image = asset;
+        this._main_caption = _ensure_cheerio(caption);
```
sgnn7: `s/_main_caption/_main_image_caption/`. If it's the generqal caption, we should have a different setter for it.

lib/index.js
```diff
@@ -170,16 +170,107 @@ class VideoAsset extends BaseAsset {
...
+            this._date_published = new Date(value);
+    }
+
+    // Takes a string or array of strings
+    set_authors(value) {
+        if (value instanceof Array)
+            this._authors = value;
+        else
+            this._authors = [value];
+    }
```
sgnn7: Much more terse and easier on the eyes:
```
this._authors = value;
if (!(value instanceof Array)) {
  value = [value];
}
```

lib/index.js
```diff
@@ -170,16 +170,107 @@ class VideoAsset extends BaseAsset {
...
         this._content_type = "text/html";
     }
 
-    set_document(value) { this._document = cheerio.load(value); }
+    // Takes a Date or something that a Date can be constructed out of
+    set_date_published(value) {
+        if (value instanceof Date)
+            this._date_published = value;
+        else
+            this._date_published = new Date(value);
```
sgnn7: You can do the same as the authors suggestion here

examples/ars.js
```diff
@@ -0,0 +1,115 @@
...
+
+        asset.render();
+        hatch.save_asset(asset);
+    });
+}
+
+function main() {
+    let hatch = new Libingester.Hatch('ars');
+    const FEED_URI = 'http://feeds.arstechnica.com/arstechnica/open-source';
+    RssToJson.load(FEED_URI)
```
sgnn7: Generally we've been keeping the first `then` on the same line as the parent.

examples/ars.js
```diff
@@ -0,0 +1,115 @@
...
+    .then(rss =>
+        Promise.all(rss.items.map(entry => ingest_article(hatch, entry))))
+    .then(() => hatch.finish())
+    .catch(err => console.log('there was an error', err));
```
sgnn7: You should re-throw the error or the caller can't tell we had a failure.

examples/ars.js
```diff
@@ -0,0 +1,115 @@
...
+    });
+}
+
+function main() {
+    let hatch = new Libingester.Hatch('ars');
+    const FEED_URI = 'http://feeds.arstechnica.com/arstechnica/open-source';
+    RssToJson.load(FEED_URI)
+    .then(rss =>
+        Promise.all(rss.items.map(entry => ingest_article(hatch, entry))))
+    .then(() => hatch.finish())
```
sgnn7: I wonder if this line could be `.then(hatch.finish)`?

examples/ars.js
```diff
@@ -0,0 +1,115 @@
...
+        asset.set_body($('section'));
+
+        asset.render();
+        hatch.save_asset(asset);
+    });
+}
+
+function main() {
+    let hatch = new Libingester.Hatch('ars');
+    const FEED_URI = 'http://feeds.arstechnica.com/arstechnica/open-source';
```
sgnn7: Move this const to top of the file

examples/ars.js
```diff
@@ -0,0 +1,115 @@
...
+const Libingester = require('libingester');
+const thenifyAll = require('thenify-all');
+const RssToJson = thenifyAll(require('rss-to-json'), {}, ['load']);
+const fs = thenifyAll(require('fs'), {}, ['writeFile']);
+
+function remove_intermediate($, selector) {
+    $(selector).each((i, elem) => $(elem).replaceWith($(elem).contents()));
+}
+
+const COMMENT_NODE = 8;
```
sgnn7: Keep consts with static vals at the top

examples/ars.js
```diff
@@ -0,0 +1,115 @@
...
+
+function ingest_article(hatch, {title, description, url, created}) {
+    return Libingester.util.fetch_html(url)
```
sgnn7: Ditto on keeping first `then` on same line as parent.

examples/ars.js
```diff
@@ -0,0 +1,115 @@
...
+        asset.set_synopsis(description);
+        asset.set_date_published(created);
+        asset.set_last_modified_date(new Date(created));
+        asset.set_source_name('Ars Technica');
+        asset.set_license('Proprietary');
+        asset.set_section('open-source');
+
+        let canonical_uri = $('head link[rel="canonical"]').attr('href');
+        asset.set_canonical_uri(canonical_uri);
+        asset.set_read_more_link(`
```
sgnn7: Hmm.. why can't this be in libingester as a wrapper on the value?

examples/ars.js
```diff
@@ -0,0 +1,115 @@
...
+</p>
+        `);
+
+        let thumb_uri = $('meta[property="og:image"]').attr('content');
+        let thumb_asset = Libingester.util.download_image(thumb_uri);
+        hatch.save_asset(thumb_asset);
+        asset.set_thumbnail(thumb_asset);
+
+        // Clean up
+        $('div').filter(function () {
```
sgnn7: Can we use a lambda here - should be a bit easier to read.

examples/ars.js
```diff
@@ -0,0 +1,115 @@
...
+        $('head, script, .site-header, .ad, #social-left').remove();
+        remove_intermediate($, '.site-wrapper');
+        remove_intermediate($, '.content-wrapper');
+        remove_intermediate($, '.column-wrapper');
+        remove_intermediate($, '.left-column');
+        remove_intermediate($, '.article-content');
+        $('.page-numbers').remove(); // oops, should deal with multiple pages
+        $('.story-sidebar, .enlarge-link, .post-upperdek').remove();
+        $('.article-author, #social-footer, #article-footer-wrap').remove();
+        $('.site-footer').remove();
```
sgnn7: Use loops for these - most seem to be the same thing just different tag/class.

examples/ars.js
```diff
@@ -0,0 +1,115 @@
...
+
+        let first_para = $('section p').first();
+        asset.set_lede(first_para);
+        $(first_para).remove();
+
+        // Clean up classes for readability
+        $('body, article, section, figure, figcaption').removeAttr('class');
+        $('article').removeAttr('itemscope');
+        $('article').removeAttr('itemtype');
+        $('figure').removeAttr('style');
```
sgnn7: Ditto - use loops

examples/ars.js
ptomato: Not sure I understand you correctly; there is no caller here, this is `main()`?

sgnn7: I _think_ that if you don't re-throw here, the program will exit with status_code of 0 which won't let the invoker know that something went wrong.

lib/index.js
ptomato: The existing metadata property is called `sourceName`, so I thought it should, but no big deal.

examples/ars.js
ptomato: I find it awful to mix promises and callbacks in the same code.

sgnn7: Good ol' JavaScript - there's usually not a nice way to do anything :)

examples/ars.js
ptomato: No, that binds a value of `this` that overrides the value that cheerio binds, so `$(this)` wouldn't work inside.

sgnn7: Cool - I didn't actually know that

examples/ars.js
ptomato: Would have to be `hatch.finish.bind(hatch)` which I find less readable - let me know if you still want it changed.

sgnn7: Leave it then - I guess I still have troubles grasping JS' `this` references in some of this flows.

examples/ars.js
ptomato: I did not manage to find out how to render a cheerio object as HTML including the root (as opposed to rendering the `innerHTML`.) So the `<p>` is actually stripped when you call `.html()` on the Cheerio object. If you know how to avoid that, it would be a pleasure to change it...

examples/ars.js
ptomato: OK, got it.

lib/index.js
ptomato: Let's look at that on Monday :smile:

sgnn7: Sounds good!

sgnn7:
@ptomato LGTM just needs the split-out of the template.

lib/index.js
```diff
@@ -170,16 +170,105 @@ class VideoAsset extends BaseAsset {
...
+
+        this._read_more_link('a').addClass('eos-show-link');
+
+        let body = `
+<header>
+    <div class="context">
+        <p id="context-link"></p>
+    </div>
+    <div class="extra-header-right">
+        <p class="date-published">${this._date_published}</p>
```
tchx84: @ptomato @sgnn7 thinking this could be a good place to use metadata API, because here we want to show the date according to the final user system settings. right?

e.g. 28/12/2017 vs 12/28/2017

lib/index.js
```diff
@@ -170,16 +171,75 @@ class VideoAsset extends BaseAsset {
...
+            };
+            if (this._main_image_caption.length) {
+                main_image['caption'] = cheerio.html(this._main_image_caption);
+            }
+        }
+
+        cheerio('a', this._read_more_link).addClass('eos-show-link');
+
+        let template_path = path.join(__dirname, './assets/news-article.mst');
+        let template = fs.readFileSync(template_path, 'utf8');
```
ptomato: Since this is an I/O operation we could consider making it async and having `render()` return a promise, but if so, we should do it in both formats.

sgnn7:
LGTM

PR #33, opened 2017-06-05
Bump to version 2.0.0

ptomato:
With the article formats being merged, this is an API break. We have
already dubbed the article formats "Libingester 2" in conversations, so
it makes sense to bump to that version.

https://phabricator.endlessm.com/T17387

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "libingester",
-  "version": "1.0.4",
+  "version": "2.0.0",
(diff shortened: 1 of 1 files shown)


ptomato:
I guess only merge this after the article formats get merged?

sgnn7:
@ptomato Yeah - it would make no sense to do that before the formats get in.

sgnn7:
CR pass since at least one template has been merged.

PR #34, opened 2017-06-08
News article stylesheet

ptomato:
https://phabricator.endlessm.com/T17401


lib/index.js
```diff
@@ -308,12 +310,44 @@ class NewsArticle extends BaseAsset {
...
         }
 
+        let quote_svg_path = path.join(__dirname, './assets/quote.svg');
+        let quote_decoration_xml = fs.readFileSync(quote_svg_path, 'utf8');
+        let quote_decoration = cheerio.load(quote_decoration_xml, {
+            xmlMode: true,
+        });
+        cheerio('blockquote', this._body).prepend(`<div class="quote-decoration">
+            ${quote_decoration.html('svg')}
+        </div>`);
```
sgnn7: I'm a little confused by this -are we assuming that everything is surrounded by quotes?

ptomato: This should add the quote decoration as the first child of every `<blockquote>` element.

sgnn7: I see what you mean - I thought it was prepending blockquote on the body with nothing on it. Thanks.

sgnn7:
:+1:

PR #37, opened 2017-06-09
previewer: Show article format specific metadata

ptomato:
Some metadata is only added in the case of a specific article format.
Show that in the previewer, but only if it is present.

### DIFF
--- a/previewer/metadata_renderer.js
+++ b/previewer/metadata_renderer.js
@@ -17,7 +17,13 @@ const visibleProps = [ 'assetID',
                        'license',
                        'tags',
                        'lastModifiedDate',
-                       'revisionTag' ]
+                       'revisionTag',
+
+                       // Properties specific to certain asset types
+                       'author',
+                       'authors',
+                       'published',
+                       'sourceName' ]
 
 $(document).ready(function(){
 })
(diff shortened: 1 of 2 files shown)


previewer/metadata_renderer.js
```diff
@@ -45,6 +51,8 @@ setMetadataAssetID = function(ID) {
     )
 
     visibleProps.forEach(function(prop) {
+      if (!asset.hasOwnProperty(prop))
+        return;
```
sgnn7: This is a tricky one - I like the idea that this panel shows all the things that could be filled in and not just what you have filled in already to help libingester devs to know what they are missing. Maybe we can have this list dynamically generate the rows based on the asset type and some consts in libingester iself?

ptomato: How were you envisioning that? I think we'd still have to have some list, because you would want the common properties to always appear in the same order.

sgnn7: @ptomato - yeah no idea about that. maybe we alway show the common ones and then if there's anything left over we append it to the list.

ptomato:
Superseded by https://github.com/endlessm/libingester/pull/38

PR #38, opened 2017-06-12
More metadata in previewer

ptomato:
Includes other minor fixes.

https://phabricator.endlessm.com/T17569

### DIFF
--- a/lib/assets/blog-stylesheet.scss
+++ b/lib/assets/blog-stylesheet.scss
@@ -1,20 +1,14 @@
 $title-font: 'Fira Sans'!default;
 $body-font: 'Merriweather' !default;
-$display-font: 'Fira Sans' !default;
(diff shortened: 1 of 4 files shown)


sgnn7:
@ptomato Looks good!

PS: Do we have a story for pulling the list of these extra field maps from the actual libingester code?

ptomato:
I don't think it should come from the libingester code, since these fields belong to the client side models. (And I'm not sure it's possible to do it automatically since you want to have a whitelist of fields that are useful to show in the previewer)

sgnn7:
I'd lean that we still want to show all possible fields that libingester dumps in metadata for a specific object type (not just strictly useful ones) and have that dynamically updated as we develop the models but it's really not a mandatory thing anyways so we can punt on this and see if anyone would like to see this feature.

PR #39, opened 2017-06-13
Dictionary object model

syndacks:
Added a "DictionaryAsset" to index.js. This will support the "DictionaryObject" data model that is for DiscoveryFeed. I also added some related tests in verify.js.

https://phabricator.endlessm.com/T17329

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -186,7 +186,7 @@ function _ensure_cheerio(value) {
 class DictionaryAsset extends BaseAsset {
     constructor() {
         super();
-        this._object_type = "DictionaryObject";
+        this._object_type = "DictionaryObjectModel";
         this._content_type = "text/html";
     }
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -191,7 +191,7 @@ class DictionaryAsset extends BaseAsset {
     }
 
     set_word (value) { this._word = value; }
-    set_word_definition (value) { this._word_definition = value; }
+    set_definition (value) { this._definition = value; }
     set_part_of_speech (value) { this._part_of_speech = value; }
     set_tags (value) { this._tags = value; }
 
@@ -199,7 +199,7 @@ class DictionaryAsset extends BaseAsset {
         const metadata = super.to_metadata();
         Object.assign(metadata, {
             'word': this._word,
-            'wordDefinition': this._word_definition,
+            'definition': this._definition,
             'partOfSpeech': this._part_of_speech,
             'tags': [this._tags],
             'document': this._document.html()
@@ -213,7 +213,7 @@ class DictionaryAsset extends BaseAsset {
 
         let document = mustache.render(template, {
             word: this._word,
-            word_definition: this._word_definition,
+            definition: this._definition,
             part_of_speech: this._part_of_speech
         });
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -183,6 +183,48 @@ function _ensure_cheerio(value) {
     return value;
 }
 
+class DictionaryAsset extends BaseAsset {
+    constructor() {
+        super();
+        this._object_type = "DictionaryObject";
+        this._content_type = "text/html";
+    }
+
+    set_word (value) { this._word = value; }
+    set_word_definition (value) { this._word_definition = value; }
+    set_part_of_speech (value) { this._part_of_speech = value; }
+    set_tags (value) { this._tags = value; }
+
+    to_metadata() {
+        const metadata = super.to_metadata();
+        Object.assign(metadata, {
+            'word': this._word,
+            'wordDefinition': this._word_definition,
+            'partOfSpeech': this._part_of_speech,
+            'tags': [this._tags],
+            'document': this._document.html()
+        });
+        return metadata;
+    }
+
+    render() {
+        let template_path = path.join(__dirname, './assets/dictionary-article.mst');
+        let template = fs.readFileSync(template_path, 'utf8');
+
+        let document = mustache.render(template, {
+            word: this._word,
+            word_definition: this._word_definition,
+            part_of_speech: this._part_of_speech
+        });
+
+        this._document = cheerio.load(document);
+    }
+
+    _make_tags() { return this._tags; }
+}
+
+exports.DictionaryAsset = DictionaryAsset;
+
 class BlogArticle extends BaseAsset {
     constructor() {
         super();
--- a/lib/verify.js
+++ b/lib/verify.js
@@ -41,7 +41,7 @@ function verify_metadata(metadata) {
         assert(!!metadata['cdnFilename'], "Image object missing cdnFilename (has no data?)");
     } else if (object_type === "DictionaryObjectModel") {
         assert(!!metadata['word'], "metadata missing word");
-        assert(!!metadata['wordDefinition'], "metadata missing wordDefinition");
+        assert(!!metadata['definition'], "metadata missing definition");
     } else {
         throw new VerificationError("metadata has wrong objectType");
     }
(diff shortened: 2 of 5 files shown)


sgnn7:
@syndacks Don't forget about https://github.com/endlessm/libingester/blob/master/previewer/metadata_renderer.js#L7

Also, squash your commits.

syndacks:
@sgnn7 Thanks, reviewed and squashed

lib/index.js
```diff
@@ -178,6 +178,35 @@ function _ensure_cheerio(value) {
...
 
+class DictionaryAsset extends BaseAsset {
+    constructor() {
+        super();
+        this._object_type = "DictionaryObject";
+        this._content_type = "text/html";
+    }
+
+    set_word (value) { this._word = value; }
+    set_wordDefinition (value) { this._wordDefinition = value; }
```
ptomato: drive-by nitpick: don't mix `camelCase` and `snake_case`

syndacks: @ptomato thanks! please continue to drive/walk/run-by any code of mine and offer review.

sgnn7: +1 on casing

lib/index.js
```diff
@@ -178,6 +178,35 @@ function _ensure_cheerio(value) {
...
+    set_partOfSpeech (value) { this._partOfSpeech = value; }
+    set_tags (value) { this._tags = value; }
+
+    to_metadata() {
+        const metadata = super.to_metadata();
+        Object.assign(metadata, {
+            'word': this._word,
+            'wordDefinition': this._wordDefinition,
+            'partOfSpeech': this._partOfSpeech,
```
sgnn7: Ditto on casing

lib/index.js
```diff
@@ -178,6 +178,35 @@ function _ensure_cheerio(value) {
...
+            'tags': [this._tags]
+        });
+        console.log(metadata);
```
sgnn7: Probably needs to be removed

lib/verify.js
```diff
@@ -30,6 +30,8 @@ function verify_metadata(metadata) {
         assert(!!metadata['document'], "metadata missing document");
     } else if (object_type === 'ImageObject') {
         assert(!!metadata['cdnFilename'], "Image object missing cdnFilename (has no data?)");
+    } else if (object_type === "DictionaryObject") {
+        assert(!!metadata['word'], "metadata missing word");
```
sgnn7: Not checking that word definition and part of speech is set as well?

previewer/metadata_renderer.js
```diff
@@ -7,7 +7,8 @@ hatchFolder = main.hatchFolder
 const extraPropsByModelType = {
   'ArticleObject': ['authors', 'published', 'sourceName'],
   'ImageObject': ['caption', 'height', 'width'],
-  'VideoObject': ['caption', 'duration', 'height', 'transcript', 'width']
+  'VideoObject': ['caption', 'duration', 'height', 'transcript', 'width'],
+  'DictionaryObject': ['word', 'wordDefinition', 'partOfSpeech']
```
sgnn7: Ditto on casing

lib/verify.js
syndacks: Good point, I'm going to leave off "part of speech" as that doesn't seem as important as a word/wordDefinition pair.

lib/index.js
```diff
@@ -178,6 +178,34 @@ function _ensure_cheerio(value) {
...
+    set_part_of_speech (value) { this._part_of_speech = value; }
+    set_tags (value) { this._tags = value; }
+
+    to_metadata() {
+        const metadata = super.to_metadata();
+        Object.assign(metadata, {
+            'word': this._word,
+            'wordDefinition': this._word_definition,
+            'partOfSpeech': this._part_of_speech,
+            'tags': [this._tags]
```
wgreenberg: this line shouldn't be necessary, also `_tags` as a variable implies it should be an array

lib/index.js
```diff
@@ -178,6 +178,34 @@ function _ensure_cheerio(value) {
     return value;
 }
 
+class DictionaryAsset extends BaseAsset {
+    constructor() {
+        super();
+        this._object_type = "DictionaryObject";
+        this._content_type = "text/html";
```
wgreenberg: I don't think this is actually correct, since afaik there's no HTML involved here. To be honest I'm not sure what the content type of a "word" is

ptomato: I might suggest having some sort of rendered HTML in the document body anyway, even if it's not used directly in the discovery feed...

lib/index.js
syndacks: I had to explicitly set the tags as an array to match our `verify` condition.

lib/index.js
syndacks: Sounds good - there's a `render` method now in the DictionaryAsset class which sets the relevant fields (word, wordDefinition, partOfSpeech) to already existing fields on the Blog template.

syndacks: @wgreenberg let me know what you think

ptomato: I would not reuse the blog template for this - just make some simple HTML document with no styling for the time being

syndacks: thanks @ptomato - changed made.

lib/index.js
```diff
@@ -183,6 +183,48 @@ function _ensure_cheerio(value) {
...
+    set_definition (value) { this._definition = value; }
+    set_part_of_speech (value) { this._part_of_speech = value; }
+    set_tags (value) { this._tags = value; }
+
+    to_metadata() {
+        const metadata = super.to_metadata();
+        Object.assign(metadata, {
+            'word': this._word,
+            'definition': this._definition,
+            'part-of-speech': this._part_of_speech,
```
wgreenberg: in the other classes, we use camelCase for metadata keys. s/part-of-speech/partOfSpeech

previewer/metadata_renderer.js
```diff
@@ -7,7 +7,8 @@ hatchFolder = main.hatchFolder
 const extraPropsByModelType = {
   'ArticleObject': ['authors', 'published', 'sourceName'],
   'ImageObject': ['caption', 'height', 'width'],
-  'VideoObject': ['caption', 'duration', 'height', 'transcript', 'width']
+  'VideoObject': ['caption', 'duration', 'height', 'transcript', 'width'],
+  'DictionaryObjectModel': ['word', 'definition', 'part-of-speech']
```
wgreenberg: nit: trailing comma

wgreenberg: also change to partOfSpeech

lib/index.js
syndacks: General question...these metadata fields need to match up with: https://github.com/endlessm/endless-content-factory/pull/742

right?

cosimoc:
Needs rebase to master.

syndacks:
@cosimoc rebased
@wgreenberg ready for review

lib/index.js
```diff
@@ -204,6 +204,48 @@ class VideoAsset extends BaseAsset {
...
+    set_word (value) { this._word = value; }
+    set_definition (value) { this._definition = value; }
+    set_part_of_speech (value) { this._part_of_speech = value; }
+    set_tags (value) { this._tags = value; }
+
+    to_metadata() {
+        const metadata = super.to_metadata();
+        Object.assign(metadata, {
+            'word': this._word,
+            'definition': this._definition,
```
wgreenberg: shouldn't this be wordDefinition?

syndacks: @wgreenberg I think @smspillaz changed this?...i'm just about to head out for the night so can take a look on Monday

smspillaz: This should be `definition` - see https://github.com/endlessm/endless-content-factory/pull/742/commits/5dc7dcdef80e9e93aecfe5f6fcb7c7d63db4336d

Unless there's something else that needs to change?

syndacks: `definition` confirmed correct

wgreenberg:
LGTM

PR #40, opened 2017-06-14
T17565

wgreenberg:


lib/util.js
```diff
@@ -7,6 +7,8 @@ const parseDataUrl = require('parse-data-url');
 const request = require('request');
 const url = require('url');
 const somaDOM = require('./soma_dom');
+const feedparser = require('feedparser');
+const moment = require('moment');
```
sgnn7: Sort imports

lib/util.js
```diff
@@ -193,3 +195,52 @@ function encode_uri(uri) {
...
+    const oldest_date = moment().subtract(max_days_old, 'days');
+
+    return _fetch_rss_page(feed_url, 1, [], max_items, oldest_date);
+}
+
+function _fetch_rss_json (feed_url) {
+    return new Promise((resolve, reject) => {
+        const req = request.get(feed_url);
+        const parser = new feedparser();
+        req.pipe(parser);
```
sgnn7: Shouldn't this be after the error callback setting?

lib/util.js
```diff
@@ -193,3 +195,52 @@ function encode_uri(uri) {
...
+// Assume a wordpress feed (https://codex.wordpress.org/Pagination#Adding_the_.22paged.22_parameter_to_a_query)
+function _fetch_rss_page (feed_url, page, items, max_items, oldest_date) {
+    let paginated_url = feed_url;
+    if (page > 1) {
+        paginated_url += `?paged=${page}`;
+    }
+
+    return _fetch_rss_json(paginated_url).then(feed_json => {
+        const recent_enough_items = feed_json.items.filter(article => article.pubdate >= oldest_date);
+        // If we've run into articles which are too old, cease crawling
```
sgnn7: Is chronological ordering guaranteed?

lib/util.js
```diff
@@ -193,3 +195,52 @@ function encode_uri(uri) {
...
+        if (recent_enough_items.length !== feed_json.items.length) {
+            return items.concat(recent_enough_items);
+        }
+
+        const new_items = items.concat(feed_json.items);
+        // If we've hit the desired number of items, also cease crawling
+        if (new_items.length >= max_items) {
+            return new_items.slice(max_items);
+        }
```
sgnn7: Why not just concat the feed list directly with slice and then check the length which would prevent the need for `new_items` var completely?

lib/util.js
wgreenberg: well, it's guaranteed to be in the order of the feed

lib/util.js
wgreenberg: shouldn't matter, since events can't be emitted synchronously

lib/util.js
wgreenberg: could you give an example? not sure I get the suggestion

lib/util.js
sgnn7: Is that a wordpress thing or just something we've seen?

lib/util.js
sgnn7: Didn't try running this but this is what I was suggesting:
```
items.concat(feed_json.slice(0, items.length - max_items));
if (items.length == max_items) {
  return items;
}
```

wgreenberg: concat is non-destructive, so we'd need a temporary variable anyhow -- IMO the more explicit (though potentially more verbose) way is clearer

sgnn7: I get your point but it's a convoluted way of getting what you want - you add everything then if it's over the max then you slice out what you need vs getting up-to the exact amount you need and if it's the max returning. I don't feel strongly enough to hold up the PR over it but imo it's much more overhead for the marginal clarity it provides.

lib/util.js
sgnn7: hmm I would think that during execution of this block here the ops are still ordered so the `pipe` connecting would precede the handler for error which might not fire if something occurs between L209 and 211/212 executing. Maybe I'm missing something?

lib/util.js
wgreenberg: fair nuff, updated the code

lib/util.js
wgreenberg: ah turns out I misunderstood how the event loop handles this, I think you're right. Interestingly, if an event emitter ever emits an `'error'` event and there aren't any listeners, it'll just throw an exception

PR #41, opened 2017-06-14
Libingester: Add language as the constructor param for Hatch

sgnn7:
Also fixed nodev5 compat and some ingesters.

Need to update wiki after completion. Also needs to stay in dev for e-c-f ingester changes to support this.

https://phabricator.endlessm.com/T17163

### DIFF
--- a/examples/ars.js
+++ b/examples/ars.js
@@ -14,16 +14,16 @@ function remove_intermediate($, selector) {
     $(selector).each((i, elem) => $(elem).replaceWith($(elem).contents()));
 }
 
-function ingest_article(hatch, {title, description, url, created}) {
-    return Libingester.util.fetch_html(url).then($ => {
-        const BASE_URI = Libingester.util.get_doc_base_uri($, url);
+function ingest_article(hatch, entry) {
+    return Libingester.util.fetch_html(entry.url).then($ => {
+        const BASE_URI = Libingester.util.get_doc_base_uri($, entry.url);
         let asset = new Libingester.NewsArticle();
 
-        console.log('processing', title);
-        asset.set_title(title);
-        asset.set_synopsis(description);
-        asset.set_date_published(created);
-        asset.set_last_modified_date(new Date(created));
+        console.log('processing', entry.title);
+        asset.set_title(entry.title);
+        asset.set_synopsis(entry.description);
+        asset.set_date_published(entry.created);
+        asset.set_last_modified_date(new Date(entry.created));
         asset.set_source('Ars Technica');
         asset.set_license('Proprietary');
         asset.set_section('open-source');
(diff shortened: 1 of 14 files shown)


examples/day_in_history.js
```diff
@@ -52,15 +52,20 @@ function ingestArticleProfile(hatch, uri) {
...
+    })
+    .catch(err => {
+        console.error(err.stack);
+        throw err;
     });
 }
 
 function main() {
-    const hatch = new libingester.Hatch();
+    const hatch = new libingester.Hatch('day_in_history', 'en', { argv: process.argv.slice(2) });
```
wgreenberg: hmm, I didn't realize this was how we were suggesting people to write hatch arguments. this seems very prone to being forgotten/messed up -- IMO argparsing should always only be handled by the "main" function, and then turned into normal function arguments elsewhere. libingester could provide this I guess

examples/ars.js
```diff
@@ -14,16 +14,16 @@ function remove_intermediate($, selector) {
     $(selector).each((i, elem) => $(elem).replaceWith($(elem).contents()));
 }
 
-function ingest_article(hatch, {title, description, url, created}) {
-    return Libingester.util.fetch_html(url).then($ => {
-        const BASE_URI = Libingester.util.get_doc_base_uri($, url);
+function ingest_article(hatch, entry) {
```
wgreenberg: alternately, you could do 
```
function ingest_article(hatch, entry) {
    const { title, description, url, created } = entry;
    return Libingester.util.fetch_html(url).then($ => { ... })
}
```

lib/index.js
```diff
@@ -435,20 +435,24 @@ function is_subset(a, b) {
...
-        // but otherwise add it to the prefix
         if (!name) {
-            name = "";
-        } else {
-            name += '_';
+            throw new Error("ERROR! Hatch name must be specified!");
+        }
+
+        // The string check is to ensure that we don't consider older-api `options`
+        // as the language which was in this position before
```
wgreenberg: why isn't `language` a member of `options`?

examples/day_in_history.js
sgnn7: The argv processing we can integrate into the libingester (I originally thought of it being used that way) but the problem is that we don't know how the program was invoked and if cli args are intended for the ingester and not the libingester. On top of this, making every ingester handle parsing of libingester-specific args seemed like a huge hurdle for writers of the code. I think we have to leave that decision on how many process args to trim off to the actual ingester but if we can think of a better way to handle this, I'm all about it.

lib/index.js
sgnn7: Language should be a hard option - we can't pack the shard without it so having it be a part of options (more of an optional map) seems like it would have caused unnecessary pain for ingester writers.

examples/ars.js
sgnn7: Fair but most of the vars were used once so it seemed like it was unnecessary. I'll change it if you want.

sgnn7:
@wgreenberg CR ready again

PR #42, opened 2017-06-14
T17615 - Design QA for blog/news article stylesheets

katelk28:
Design QA for blog & news article stylesheets.

Remaining question: I'd like for the blog articles to also be able to have the large quote SVG when there's a `<blockquote>`, same as the news, but am not familiar enough with libingester to be able to test if it worked. @tchx84 could you help me with this?

### DIFF
--- a/lib/assets/blog-article.mst
+++ b/lib/assets/blog-article.mst
@@ -5,9 +5,9 @@
 </head>
 
 <body>
-<header class="extra-header">
-    <div class="categories" id="categories-links"></div>
-    <div class="extra-header-right">
+<header class="article-information">
+    <div class="article-categories" id="article-categories"></div>
+    <div class="article-support-info">
         <span class="author">{{ author }}</span>
         &mdash;
         <span class="date-published" id="published-id"></span>
@@ -49,7 +49,7 @@
                 .map(set => '<a class="eos-show-link" href="' + set.id + '">' + set.title + '</a>')
                 .join(separator);
         }
-        links_helper('categories-links', true, ', ');
+        links_helper('article-categories', true, ', ');
         links_helper('tags-links', false, ' ');
 
         let published_date = new Date(metadata.published);
--- a/lib/assets/blog-stylesheet.scss
+++ b/lib/assets/blog-stylesheet.scss
@@ -1,8 +1,9 @@
-$title-font: 'Fira Sans'!default;
 $body-font: 'Merriweather' !default;
+$title-font: 'Fira Sans' !default;
 $context-font: 'Fira Sans' !default;
 $support-font: 'Fira Sans' !default;
 $body-font-composite: 'Roboto' !default;
+$title-font-composite: 'Roboto' !default;
 
 $primary-light-color: #5e7790 !default;
 $primary-dark-color: #1b242e !default;
@@ -10,7 +11,7 @@ $accent-light-color: #ff6835 !default;
 $accent-dark-color: #cc532b !default;
 $background-light-color: #e9e9e3 !default;
 
-$black60: lighten(black, 60);
+$black60: lighten(black, 100 - 60);
 $body-line-height: 1.5;
 $supplementary-height: 600px;
 
@@ -35,7 +36,7 @@ body {
     background-color: white;
     margin-left: 10vw;
     margin-right: 30vw;
-    padding-top: 2rem;
+    padding-top: 66px;
     padding-bottom: $supplementary-height;
 }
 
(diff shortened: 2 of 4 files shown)


lib/assets/news-stylesheet.scss
```diff
@@ -94,42 +96,66 @@ h1, h2, h3, h4, h5, h6 {
...
 hr {
-    border-color: $background-light-color;
-    border-width: 1px 0 0 0;
-    margin-bottom: 2rem;
+    display: block;
+    height: 1px;
+    border: 0;
+    border-top: 1px solid $background-light-color;
+    padding: 0;
+    margin: 3rem 0 2rem;
```
ptomato: The left and right margins are set in `responsive-elements` so it would be better to specify only `margin-top` and `margin-bottom` here.

lib/assets/news-stylesheet.scss
```diff
@@ -94,42 +96,66 @@ h1, h2, h3, h4, h5, h6 {
...
+    line-height: $body-line-height;
     font-weight: normal;
+    font-style: italic;
 }
 
 hr {
-    border-color: $background-light-color;
-    border-width: 1px 0 0 0;
-    margin-bottom: 2rem;
+    display: block;
```
ptomato: I think `<hr>` is `display: block` by default?

lib/assets/news-stylesheet.scss
```diff
@@ -181,6 +207,7 @@ footer {
     color: $black60;
     font-style: italic;
     line-height: 1.5;
+    margin: 1.5rem 0;
```
ptomato: Likewise, use `margin-top` and `margin-bottom` here

lib/assets/blog-article.mst
```diff
@@ -5,9 +5,9 @@
 </head>
 
 <body>
-<header class="extra-header">
```
ptomato: Why the renames?

In any case, you'll need to update the `<script>` elements in these files since they use the IDs.

lib/assets/blog-stylesheet.scss
```diff
@@ -1,16 +1,17 @@
...
+$title-font-composite: 'Roboto' !default;
 
 $primary-light-color: #5e7790 !default;
 $primary-dark-color: #1b242e !default;
 $accent-light-color: #ff6835 !default;
 $accent-dark-color: #cc532b !default;
 $background-light-color: #e9e9e3 !default;
 
-$black60: lighten(black, 60);
+$black60: transparentize(black, 1 - 0.60);
```
katelk28: Yeah, you're right now that I think about it; the specs do say to lower the opacity, but I don't really see why it's necessary for these article pages. We can go with `lighten()`. But, it's used the wrong way originally, I'll change it to `lighten(black, 100 - 60)`

lib/assets/blog-stylesheet.scss
```diff
@@ -88,33 +87,53 @@ h1, h2, h3, h4, h5, h6 {
...
     font-weight: bold;
-    line-height: 1.8;
+    line-height: 1.25;
+    margin-top: 3.0rem;
+    margin-bottom: 0.5rem;
 }
 
 h1 {
     font-size: 2.8rem;
+    + h2, + h3, + h4, + h5, + h6 {
```
katelk28: you are right! overlooked this

lib/assets/blog-article.mst
katelk28: Thought `extra-header` and `extra-header-right` don't really define the information very well and seems too specific. If a designer were looking at this, they'd be prepared to look for something having to do with "context" and "support", etc.

ptomato: OK, makes sense. I don't feel strongly one way or the other. Just always watch out with renames and use `git grep extra-header` to make sure nothing else (at least in the same git repository) is depending on the old name.

BTW, for future reference this would be a good candidate for splitting out into its own commit so you can explain in the commit message why the rename was done. No need to go back and change it now though.

katelk28: Yeah, missing the extra uses of the name was sloppy on my part, got all of them now.
And noted! thanks for the advice. But, in that commit, would it make sense to include the renaming in the CSS Files in that commit as well? Or just keep the CSS changes in one commit and the .mst file in the commit you're talking about?

ptomato: Changing the name everywhere, in one commit, and the unrelated CSS changes in another commit.

PR #43, opened 2017-06-20
Rework RSS crawl utility to handle arbitrary pagination

wgreenberg:
Since we support more than just wordpress sites, this new API will allow
the user of libingester to provide an arbitrary "paginator" function
which returns the Nth page of an RSS feed.


ptomato:
Looks fine, might be nice to provide a util API that returns a Wordpress paginator function given a feed URI?

wgreenberg:
updated

ptomato:
:+1:

PR #46, opened 2017-07-03
fix bug w/ limiting number of RSS entries

wgreenberg:


lib/util.js
```diff
@@ -234,18 +234,16 @@ function _fetch_rss_page (feed, page, items, max_items, oldest_date) {
...
 
     return _fetch_rss_json(feed_url).then(feed_json => {
-        const recent_enough_items = feed_json.items.filter(article => article.pubdate >= oldest_date);
-        // If we've run into articles which are too old, cease crawling
+        const limited_items = feed_json.items.slice(0, max_items - items.length);
+        const recent_enough_items = limited_items.filter(article => article.pubdate >= oldest_date);
+
+        // If we've run into articles which are too old, or we've hit the max
+        // number of items, cease crawling
+        const new_items = items.concat(recent_enough_items);
```
sgnn7: Move this below this check since on the last loop you're evaluating it needlessly

sgnn7:
CR pass

PR #51, opened 2017-07-14
fetch_html: allow for content-type overrides

wgreenberg:
https://phabricator.endlessm.com/T18105

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -29,13 +29,22 @@ function rp (options) {
     });
 }
 
-function fetch_html(uri) {
+function fetch_html(uri, encoding) {
     uri = encode_uri(uri);
 
-    return rp({
+    const options = {
         uri: uri,
         gzip: true,
-    }).then(res => {
+    };
+
+    // Allow user to override encoding for instances where a site isn't giving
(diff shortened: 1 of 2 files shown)


lib/util.js
```diff
@@ -29,13 +29,21 @@ function rp (options) {
...
         uri: uri,
         gzip: true,
-    }).then(res => {
+    };
+
+    // Allow user to override encoding for instances where a site isn't giving
+    // the proper content-type
+    if (encoding) {
+        options.encoding = encoding;
+    }
```
sgnn7: Since it's defaulted to null in the param list, can't you just set encoding in the options above to:

```
const options = {
  uri: uri,
  encoding: encoding,
  gzip: true,
};
```

wgreenberg: no, because `encoding: null` is meaningful to request and not what we want by default

sgnn7: K. Can you add a comment for it?

sgnn7:
@wgreenberg This will need a rebase too

wgreenberg:
fixed!

sgnn7:
CR Pass

PR #56, opened 2017-07-24
Correctly left-align .article-content

katelk28:
The `.article-content` div moves with window expansion, but only when there
is not enough content to horizontally expand the div. This also causes
`.article-support` to expand.

Keep .article-support static, and left-align `.article-content`.

https://phabricator.endlessm.com/T18513


wgreenberg:
CR pass!

PR #58, opened 2017-07-25
gallery: Add a temoralCoverage field

smspillaz:
This is used to specify the dates that the content might refer to
as opposed to the publication date of the article itself. The difference
is that the content dates should not be used for formatting, but they
might be useful for display.

https://phabricator.endlessm.com/T17886

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -329,6 +329,7 @@ class GalleryImageArticle extends BaseAsset {
         Object.assign(metadata, {
             'document': this._document.html(),
             'authors': [this._author],
+            'published': this._date_published.toISOString(),
             'contentDates': this._content_dates || []
         });
         return metadata;
--- a/lib/index.js
+++ b/lib/index.js
@@ -314,14 +314,19 @@ class GalleryImageArticle extends BaseAsset {
         }
     }
 
-    set_content_dates(value) {
-        this._content_dates = value.map((v) => {
+    // temporal_coverage here is an array of dates, sorted
+    // in ascending order. The consumer is responsible for
+    // translating this into something sensible in accordance
+    // with the user's localisation preferences per
+    // http://schema.org/temporalCoverage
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -314,12 +314,22 @@ class GalleryImageArticle extends BaseAsset {
...
+            return v;
+        });
+    }
+
     to_metadata() {
         const metadata = super.to_metadata();
         Object.assign(metadata, {
             'document': this._document.html(),
             'authors': [this._author],
-            'published': this._date_published.toISOString(),
```
sgnn7: Hmm.. I would think that the article still needs a published date

smspillaz: Is it possible to make that null?

The reason I ask is that it looks like on the website there is no date indicating when the article itself was published. I wouldn't want to use the content date, since otherwise it would get used for sorting.

smspillaz: I'm guessing that if we *have* to set a date, we can set it to the current date, although you lose deterministic sorting.

lib/index.js
```diff
@@ -314,12 +314,28 @@ class GalleryImageArticle extends BaseAsset {
...
+        }).sort();
+    }
+
     to_metadata() {
         const metadata = super.to_metadata();
         Object.assign(metadata, {
             'document': this._document.html(),
             'authors': [this._author],
             'published': this._date_published.toISOString(),
+            'temporalCoverage': this._temporal_coverage || []
```
cosimoc: Can you not instead make sure that `set_temporal_coverage()` always sets an array?

smspillaz: The issue here is that `this._temporal_coverage` is not mandatory and may be undefined. In that case we should set it to an empty array. I'll set the default value in the constructor nonetheless.

wgreenberg:
LGTM

PR #62, opened 2017-08-01
If no pubdate is found, just return a reasonable # of articles

wgreenberg:


lib/util.js
```diff
@@ -251,6 +259,15 @@ function _fetch_rss_page (feed, page, items, max_items, oldest_date) {
     }
 
     return _fetch_rss_json(feed_url).then(feed_json => {
+        // First check that the articles even have a published date -- if not,
+        // return a realistic (finite) number of articles
+        if (!_feed_has_pubdate(feed_json)) {
+            const realistic_limit = Math.min(max_items, 100);
```
sgnn7: Extract 100 to a const up top

lib/util.js
```diff
@@ -240,6 +240,14 @@ function _fetch_rss_json (feed_url) {
     });
 }
 
+function _feed_has_pubdate (feed) {
+    if (feed.items.length) [
```
sgnn7: Bracket?

lib/util.js
```diff
@@ -240,6 +240,14 @@ function _fetch_rss_json (feed_url) {
...
+        return (feed.items[0].pubdate !== null);
```
sgnn7: Parens here are an overkill probably

sgnn7: Can't we just return `feed.items.length && feed.items[0].pubdate !== null`?

lib/util.js
wgreenberg: derp

lib/util.js
wgreenberg: we can, I just think it's much clearer to spell it out as an if statement

wgreenberg: on second thought, I think your way is more clear

sgnn7:
Needs a rebase too

manuq:
Is there any case of feeds containing articles without a published date?

wgreenberg:
@manuq this was quite a while ago, but if I recall correctly we had one RSS feed which wasn't supplying the pubDate metadata

PR #63, opened 2017-08-10
T18642 (sanitize title and synopsis)

tchx84:
https://phabricator.endlessm.com/T18642

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -51,8 +51,8 @@ class BaseAsset {
 
     get asset_id() { return this._asset_id; }
 
-    set_title(value) { this._title = value; }
-    set_synopsis(value) { this._synopsis = value; }
+    set_title(value) { this._title = value.trim(); }
+    set_synopsis(value) { this._synopsis = value.trim(); }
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -51,8 +57,8 @@ class BaseAsset {
 
     get asset_id() { return this._asset_id; }
 
-    set_title(value) { this._title = value; }
-    set_synopsis(value) { this._synopsis = value; }
+    set_title(value) { this._title = _sanitize_text(value); }
```
ptomato: There is a method in the standard library for this, `value.trim()`.

tchx84: great, I'd like to keep the _sanitize_text function though, in case we want to add more things later to the sanitation. Any strong opinions? @ptomato @wgreenberg

wgreenberg: @tchx84 it's a bit tricky, since we may want to sanitize text differently for different fields. i.e. newlines aren't desired in titles, but could be desired elsewhere

tchx84: ok, I will move it inline then (for the record, I though that in the long term having trim/replace all scattered around wouldn't read too nicely).

PR #65, opened 2017-08-14
Fix bootstrap dep issue

sgnn7:
New bootstrap doesn't specify jQ as a direct dep but as a peerDependecy so this is the workaround.


syndacks:
Looks good to me, tested locally :+1:

PR #66, opened 2017-08-15
Libingester: Fix wrapping problem for article-support in Gallery Articles

katelk28:
The `.article-support` div that holds the original URI is too narrow.
- I gave it a little more width: 15vw to 20vw
- when there's nothing in the body of the content, the width of `.article-support` should be 100%, and the `.article-content` div is removed.

https://phabricator.endlessm.com/T18790

### DIFF
--- a/lib/assets/gallery-image-article.mst
+++ b/lib/assets/gallery-image-article.mst
@@ -52,17 +52,15 @@
 let categories = document.getElementById('article-categories');
 if (categories != null) {
     if (categories.children.length < 1) {
-        document.getElementById('article-title').className += ' no-margin-top';
+        document.getElementById('article-title').classList.add('no-margin-top');
     }
 }
 
 let article_content = document.getElementById('article-content');
-if (article_content != null) {
-    if (article_content.children.length < 1) {
-        document.getElementById('article-title').className += ' less-margin-bottom';
-        document.getElementById('article-support').className += ' width-full';
-        article_content.parentNode.removeChild(article_content);
-    }
+if (article_content && article_content.children.length < 1) {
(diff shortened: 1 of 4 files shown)


lib/assets/gallery-video-article.mst
```diff
@@ -62,6 +62,8 @@ let article_content = document.getElementById('article-content');
 if (article_content != null) {
     if (article_content.children.length < 1) {
```
sgnn7: I know it's not directly part of this PR but these two ifs should be combined:
`if (article_content && article_content.children.length) {`

sgnn7: Ditto on the categories lines above

lib/assets/gallery-video-article.mst
```diff
@@ -62,6 +62,8 @@ let article_content = document.getElementById('article-content');
...
         document.getElementById('article-title').className += ' less-margin-bottom';
+        document.getElementById('article-support').className += ' width-full';
```
sgnn7: Why not use toggle?
https://developer.mozilla.org/en-US/docs/Web/API/Element/classList -> toggle() method

lib/assets/gallery-image-article.mst
```diff
@@ -60,6 +60,8 @@ let article_content = document.getElementById('article-content');
 if (article_content != null) {
     if (article_content.children.length < 1) {
         document.getElementById('article-title').className += ' less-margin-bottom';
+        document.getElementById('article-support').className += ' width-full';
```
sgnn7: Ditto on squashing the `if`s and using `toggle()`

sgnn7:
@katelk28 CR Pass :+1:

PR #67, opened 2017-08-15
Libingester: Gallery Articles: article-support div full-width

katelk28:
- Fixed the incorrect div class selection when article-content is empty (should be article-support, not article-title)

https://phabricator.endlessm.com/T18790


sgnn7:
CR Pass

PR #68, opened 2017-08-15
Don't warn when output dir exists

wgreenberg:
The vast majority of the time libingester is invoked (i.e. on SOMA by
libingester-worker) the dir is created, giving spurious warnings on each
run


sgnn7:
CR Pass

PR #69, opened 2017-08-16
GalleryVideoArticle: Add play button overlay asset

katelk28:
Added a play button overlay for the video screenshot on the GalleryVideoArticle page.
Also gave the body element a min-width of 300px; when the
window shrinks, the play button becomes too large.

https://phabricator.endlessm.com/T18777


sgnn7:
@katelk28 Don't forget to bump the package.json version

ptomato:
This is something that I'd often forget as well and can be automated; how about a Travis CI script that checks that the package.json version is bumped for every pull request?

lib/assets/gallery-video-article.mst
```diff
@@ -12,6 +12,9 @@
             <figure class="main-image">
                 <img data-libingester-asset-id="{{ main_image }}">
             </figure>
+            <div class="play-button-wrapper">
+                <img src="play-button.svg" />
```
sgnn7: Hmm... I don't this path available in the built shard in this manner. I think you might need to post-process the file similarly to the quote image: https://github.com/endlessm/libingester/blob/4930408351bff245c808f2a7e2bfcdef4ecd644f/lib/index.js#L574-L581

@wgreenberg Any thoughts?

sgnn7:
@ptomato That would be a good thing to add for sure

lib/assets/gallery-video-article.mst
wgreenberg: right, the svg will have to be inline in the HTML

wgreenberg: there's not currently a way to include extra files outside of the metadata/data

sgnn7:
@katelk28 Looks good!

PR #74, opened 2017-08-23
Dictionary object changes

syndacks:
I overlooked a few methods that DictionaryAsset should probablly have.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -233,16 +233,51 @@ class DictionaryAsset extends BaseAsset {
     }
 
     render() {
-        let template_path = path.join(__dirname, './assets/dictionary-article.mst');
-        let template = fs.readFileSync(template_path, 'utf8');
+        let main_image = false;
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -229,16 +233,50 @@ class DictionaryAsset extends BaseAsset {
...
+                        };
+                    }
+                    return null;  // default handler
+                },
+            });
+        } else {
+            stylesheet = sass.renderSync({
+                file: dictionary_stylesheet,
+                outputStyle: 'compressed',
+            });
```
sgnn7: You should rework this to be clearer code:
```
// set defaults to sassOptions object (i.e.everything you do in the else part
// then if custom css, change some of those options as needed
// finally, render the stylesheet
stylesheet = sass.renderSync(sassOptions);
```

syndacks:
looks good @sgnn7 , tested locally and CR Pass

PR #75, opened 2017-08-23
Update word_of_day ingester with DictionaryAsset

syndacks:
This word_of_day ingester is nearly identical to the quote_of_day ingester, save for the DictrionaryAsset class  and the expected changes in Cheerio CSS selectors.


wgreenberg:
LGTM

PR #77, opened 2017-09-06
T19079

wgreenberg:
https://phabricator.endlessm.com/T19079

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -140,6 +140,8 @@ class BaseAsset {
                 promises.push(fs.writeFile(path.join(hatch_path, `${this._asset_id}.data`), data));
 
             return Promise.all(promises);
+        }).catch(err => {
+            hatch._failed_assets.push(this);
         }));
     }
 
@@ -161,6 +163,8 @@ class BaseAsset {
         return assets;
     }
 }
+
+exports.BaseAsset = BaseAsset;
 
 class ImageAsset extends BaseAsset {
     constructor() {
@@ -765,6 +769,7 @@ class Hatch {
     constructor(name, language, options) {
         this._promises = [];
         this._assets = [];
+        this._failed_assets = [];
         this._videos = [];
 
         options = options || {};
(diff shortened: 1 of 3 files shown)


test/lib/index.test.js
```diff
@@ -2,15 +2,50 @@
...
+        if (this._err) {
+            return Promise.reject(this._err);
+        } else {
+            return Promise.resolve();
+        }
+    }
+}
+
+function rmrf (dir) {
+    fs.readdirSync(dir).forEach(file => fs.unlinkSync(`${dir}/${file}`));
```
sgnn7: `fs-extra` could be easier but not a blocker (https://github.com/jprichardson/node-fs-extra/blob/master/lib/remove/__tests__/remove-dir.test.js)

test/lib/index.test.js
```diff
@@ -2,15 +2,50 @@
...
+        if (this._dependents) {
+            return this._dependents.map(asset => asset.asset_id);
+        } else {
+            return [];
+        }
+    }
+    _process() {
+        if (this._err) {
+            return Promise.reject(this._err);
+        } else {
```
sgnn7: No need for else

test/lib/index.test.js
```diff
@@ -2,15 +2,50 @@
...
+    },
+});
+
+class MockAsset extends libingester.BaseAsset {
+    fails_with_error(err) { this._err = err; }
+    set_dependent_assets(dependents) { this._dependents = dependents; }
+    get_dependent_assets() {
+        if (this._dependents) {
+            return this._dependents.map(asset => asset.asset_id);
+        } else {
```
sgnn7: No need for else

lib/index.js
```diff
@@ -824,6 +830,47 @@ class Hatch {
...
+
+        // In a second pass, remove all otherwise successful assets which had
+        // failed siblings
+        successful_assets = successful_assets.filter(asset => {
+            return !dependents_to_prune.includes(asset.asset_id);
+        });
+
+        // If this hatch had 100% failures, consider it failed and refuse to
+        // continue
+        if (successful_assets.length === 0 && this._assets.length > 0) {
```
sgnn7: Maybe this can be set from the caller rather than assuming that any % of failures is fine?

wgreenberg: maybe if that turns out to be necessary, but tbh I don't really see a use case for it

test/lib/index.test.js
wgreenberg: IMO the fewer dependencies the better

wgreenberg:
@sgnn7 fixed up

lib/index.js
sgnn7: I think there's somewhat of an issue if 99% of articles fail - maybe we can even hard-code some sensible level here?

wgreenberg:
@sgnn7 added the max fail rate

sgnn7:
@wgreenberg LGTM - just needs an addition of version bump

wgreenberg:
@sgnn7 done!

sgnn7:
CR Pass

PR #83, opened 2017-09-19
datePublished => published

wgreenberg:
Seems we've had the wrong name for this field set all along.

https://phabricator.endlessm.com/T19297

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "libingester",
-  "version": "2.2.22",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "libingester",
-  "version": "2.2.22",
+  "version": "2.2.23",
```
sgnn7: Needs bump to 24

sgnn7:
CR Pass

PR #84, opened 2017-09-26
Rework GalleryVideoArticle main images

wgreenberg:
A GVA can now either provide a main_image, in which case it'll be
featured, or it can omit one, in which case a thumbnail will be
generated. In either case, the same play button SVG will be overlaid on
top of the video thumbnail as is done in other templates.

https://phabricator.endlessm.com/T19409


sgnn7:
CR Pass

PR #85, opened 2017-09-29
Article pages: Create partials that contain shared CSS properties

katelk28:
Article pages: Create partials that contain shared CSS properties. Read commit for more information.
I put this all in 1 commit because there are a few moving parts that had to be done in one go.
Design QA was done by me.

https://phabricator.endlessm.com/T19470


katelk28:
@ptomato Awesome! will do.

PR #92, opened 2017-11-15
Allow RSS enty values to be overridden

wgreenberg:
Since parameterizing these over the command line would necessitate
manual changes to every single ingester script, let's just use the
environment.

https://phabricator.endlessm.com/T19959

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -227,6 +227,18 @@ function encode_uri(uri) {
 exports.encode_uri = encode_uri;
 
 function fetch_rss_entries (feed, max_items=Infinity, max_days_old=1) {
+    // Allow environment variables to override this
+    try {
+        if (process.env.LIBINGESTER_MAX_ITEMS) {
+            max_items = parseInt(process.env.LIBINGESTER_MAX_ITEMS);
+        }
+        if (process.env.LIBINGESTER_MAX_DAYS_OLD) {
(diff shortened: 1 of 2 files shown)


lib/util.js
```diff
@@ -227,6 +227,18 @@ function encode_uri(uri) {
...
+    try {
+        if (process.env.LIBINGESTER_MAX_ITEMS) {
+            max_items = parseInt(process.env.LIBINGESTER_MAX_ITEMS);
+        }
+        if (process.env.LIBINGESTER_MAX_DAYS_OLD) {
+            max_days_old = parseInt(process.env.LIBINGESTER_MAX_DAYS_OLD);
+        }
+    } catch (err) {
+        throw new Error(`Couldn't parse RSS entry overrides: ${err}`);
+    }
```
sgnn7: I think you can do this without try/catch:
```
node
> host = "blah"
'blah'
> process.env['MISSING']
undefined
> process.env['MISSING'] || host
'blah'
> process.env['HOST'] || host
'localhost
```

sgnn7: Edit: maybe not the parsing I suppse

wgreenberg: yeah, the parsing was why I wrapped them

sgnn7:
One small comment but not a blocker. CR pass.

PR #94, opened 2017-11-20
Fix GalleryVideo header image size restriction

katelk28:
Restricts the header image in the gallery video article page from becoming too tall (particularly if the preview image is too tall). Design QA done by me.

Missed this in my last PR done on the header image CSS.

https://phabricator.endlessm.com/T20039


wgreenberg:
CR pass

PR #95, opened 2017-11-25
Content body: remove target _blank from links

manuq:
https://phabricator.endlessm.com/T19357

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -47,11 +47,11 @@ function getTimestamp() {
     return year + month + day + "_" +  hour + minute + second;
 }
 
-function _clean_body(body) {
+function _fix_links(content) {
     // External links with target="_blank" won't open in default
     // browser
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -47,6 +47,13 @@ function getTimestamp() {
     return year + month + day + "_" +  hour + minute + second;
 }
 
+function _clean_body(body) {
+    // External links with target="_blank" won't open in default
+    // browser
+    cheerio('a[target="_blank"]', body).removeAttr('target');
+    return body;
+}
+
```
sgnn7: @manuq I think there's already a method to do this in libingester iirc and we keep adding this to individual ingesters - can we move this functionality fully to libingester?

ptomato:
Needs a version bump in package.json, then it's good to merge I think.

manuq:
I found one more case in a ingester, for NewsArticle.set_lede(). So I renamed the new function and wrapped set_lede() as well.

PR #97, opened 2017-11-30
gallery: Make img selector more specific

ptomato:
The previous selector would apply the rule to the social sharing icons,
which would mess up their padding because they were given a width of 100%.

https://phabricator.endlessm.com/T20019


katelk28:
CR pass!

PR #100, opened 2017-12-07
Allow to set tags to any asset

manuq:
Unify the API across all different asset types. The exception is
NewsArticle, for which we only allow one tag as the section.

https://phabricator.endlessm.com/T20295


cosimoc:
LGTM, but perhaps @sgnn7 or @wgreenberg want to have another look.

wgreenberg:
Cr pass

PR #103, opened 2017-12-14
hatch-previewer: reset the document preview scroll to top on document…

manuq:
… switch

https://phabricator.endlessm.com/T20476


sgnn7:
Good catch!

PR #107, opened 2018-01-03
Retry

manuq:


lib/util.js
```diff
@@ -19,22 +20,42 @@ const requestDefaults = {
...
 // Wrap request results in a promise
-function rp (options) {
-    return new Promise((resolve, reject) => {
-        pooledRequest(options, (err, res, body) => {
-            if (err) return reject(err);
-            resolve(res);
-        });
+let rp = promisify(request.defaults(requestDefaults));
+
+function wait (timeout) {
```
sgnn7: I'd rename this to something more obvious like "delayExecution".

lib/util.js
```diff
@@ -19,22 +20,42 @@ const requestDefaults = {
...
+const RETRY_START_MS = 800;
 
 // Wrap request results in a promise
-function rp (options) {
-    return new Promise((resolve, reject) => {
-        pooledRequest(options, (err, res, body) => {
-            if (err) return reject(err);
-            resolve(res);
-        });
+let rp = promisify(request.defaults(requestDefaults));
```
sgnn7: We generally try to avoid promisifying since it makes the API harder to understand. On top of this, there are modules that already do this: https://github.com/request/request#promises--asyncawait

sgnn7: PS: This should be a const

lib/util.js
```diff
@@ -19,22 +20,42 @@ const requestDefaults = {
...
+            resolve()
+        }, timeout);
     });
 }
 
+
+async function requestRetrying (options) {
+    for (let i = 0; i < MAX_RETRIES; i++) {
+        try {
+            return await rp(options);
```
sgnn7: We are on Node.js 6 as our base version which AFAIK doesn't support native await keyword

lib/util.js
```diff
@@ -19,22 +20,42 @@ const requestDefaults = {
...
+        } catch (err) {
+            if (err.code == 'ECONNRESET' || err.code == 'ETIMEDOUT') {
+                const timeout = Math.pow(2, i) * RETRY_START_MS;
+                await wait(timeout);
```
sgnn7: Ditto on await

lib/util.js
```diff
@@ -19,22 +20,42 @@ const requestDefaults = {
         'User-Agent': USER_AGENT,
     },
 };
-let pooledRequest = request.defaults(requestDefaults);
+
+// Exponential wait: 800ms, 1600ms, 3200ms
+const MAX_RETRIES = 3;
+const RETRY_START_MS = 800;
```
sgnn7: Confusing name - maybe RETRY_BACKOFF_MS would be a bit better

lib/util.js
```diff
@@ -19,22 +20,42 @@ const requestDefaults = {
...
+async function requestRetrying (options) {
+    for (let i = 0; i < MAX_RETRIES; i++) {
+        try {
+            return await rp(options);
+        } catch (err) {
+            if (err.code == 'ECONNRESET' || err.code == 'ETIMEDOUT') {
+                const timeout = Math.pow(2, i) * RETRY_START_MS;
+                await wait(timeout);
+            } else {
+                throw err;
```
sgnn7: I'd probably swap this block around to make it less messy:
```
if (some error condition) {
  throw error
}

return things
```

lib/util.js
sgnn7: Given that we don't have await, maybe this needs to be done with promise chains rather than loop too

lib/util.js
```diff
@@ -19,22 +20,42 @@ const requestDefaults = {
...
+            }
+        }
+    }
+}
+
 function set_user_agent(userAgent) {
     const newDefaults = Object.freeze(requestDefaults);
     newDefaults.headers['User-Agent'] = userAgent;
-    pooledRequest = request.defaults(newDefaults);
+    rp = promisify(request.defaults(newDefaults));
```
sgnn7: Ditto on not promisifying

package.json
```diff
@@ -25,6 +25,7 @@
     "tar": "^4.2.0",
     "tether": "^1.4.0",
     "thenify-all": "^1.6.0",
+    "util.promisify": "^1.0.0",
```
sgnn7: Ditto on not promisifying

lib/util.js
```diff
@@ -14,23 +14,62 @@ const iconv = require('iconv');
...
 
+function delayExecution (timeout) {
+    return new Promise(resolve => {
+        setTimeout(resolve, timeout);
+    });
+}
+
+function fetchWithRetry (options, attempt = 1) {
+    return promisifiedRequest(options).then(result => {
+        return result;
```
wgreenberg: nit: this `then` can be removed

lib/util.js
```diff
@@ -14,23 +14,62 @@ const iconv = require('iconv');
...
+
+        if (err.code == 'ECONNRESET' || err.code == 'ETIMEDOUT') {
+            const timeout = Math.pow(2, attempt) * RETRY_BACKOFF_DELAY;
+            console.warn(`Delaying execution of ${options.uri} by ${timeout}`);
+            return delayExecution(timeout).then(() => {
+                return fetchWithRetry(options, attempt + 1);
+            });
+        }
+
+        // We reach here if unexpected error or we go over MAX_RETRIES
```
wgreenberg: nit: comment isn't quite correct, this only happens if it's an unexpected error -- exceeding retries happens above

wgreenberg:
CR pass

PR #109, opened 2018-01-25
index: add setter for sequenceNumber metadata

manuq:
There are now EKN templates able to sort sets and articles using the
sequenceNumber metadata.  This is useful for static ingesters.  For
example, to create videos lesson apps.

Example app already using this metadata:
https://github.com/endlessm/eos-subscriptions-apps/blob/master/com.endlessm.math.vi/content/db.json


sgnn7:
LGTM

PR #113, opened 2018-01-27
T19225 news

katelk28:
Inserting sharing feature placeholder to News. Includes redesigns of News article page.

(Most design QA has been done by me, but a second pair of eyes for glaring errors would be appreciated.)

https://phabricator.endlessm.com/T19225


lib/assets/news-stylesheet.scss
```diff
@@ -115,86 +86,91 @@ footer {
...
         }
     }
-}
 
-@media (min-width: $spacious-min-width) {
-    @include responsive-elements($body-size: 24px);
-}
-
-// Composite resolutions
-@media (device-width: 720px) and (device-height: 480px),
```
ptomato: Removing these will default composite resolutions to the `$compact-max-width` rules, which will give the body left and right margins of 15vw. Is that intentional?

lib/assets/news-stylesheet.scss
```diff
@@ -28,41 +20,23 @@ figure {
...
-    display: flex;
-    flex-wrap: wrap;
     font-family: $support-font;
     font-size: 1.5rem;
     font-style: italic;
     line-height: 1.333;
     margin-top: 2rem;
     margin-bottom: 2rem;
-
-    .quote-decoration {
```
ptomato: I'm glad you got rid of this.

lib/assets/news-stylesheet.scss
katelk28: @ptomato Yes, that's intentional. 15vw at that screen size is small enough to not take up too much space, but large enough to create good visual margins + leave room for the sharing feature on the right side.

katelk28: @ptomato good catch -- I thought I had fixed that but didn't thoroughly test composite. Just added new styling for composite mode.

PR #114, opened 2018-01-29
util: Use iconv-lite instead of node-iconv

ptomato:
iconv-lite is possible to install without having to build native code.

Minor version bump instead of micro, because it's possible that an
ingester might be using an esoteric encoding that isn't supported by
iconv-lite.

https://phabricator.endlessm.com/T20985

### DIFF
--- a/test/lib/util.test.js
+++ b/test/lib/util.test.js
@@ -158,14 +158,15 @@ describe('download_img', function() {
 // FIXME this text intermittently fails because it reaches out to an external
 // server. We should replace it with gzipped content served locally
 describe('fetch_html', () => {
-    it('can handle gzipped responses', () => {
+    it('can handle gzipped responses', done => {
         const test_url = 'https://www.kapanlagi.com/' +
                         'intermezzone/' +
                         'bule-amerika-ini-nyoba-makan-buah-duku-ekspresinya-nggak-nahan-aee243.html';
         const doctype = '<!DOCTYPE html>';
         return util.fetch_html(test_url).then((result) => {
             expect(result.html().substring(0, doctype.length)).to.equal(doctype);
-        });
+        })
+        .finally(done);
     });
 });
 
--- a/test/lib/util.test.js
+++ b/test/lib/util.test.js
@@ -155,14 +155,23 @@ describe('download_img', function() {
     });
 });
 
-// FIXME this text intermittently fails because it reaches out to an external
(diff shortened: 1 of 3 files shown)


test/lib/util.test.js
```diff
@@ -155,14 +155,22 @@ describe('download_img', function() {
...
 });
 
-// FIXME this text intermittently fails because it reaches out to an external
-// server. We should replace it with gzipped content served locally
+// FIXME these tests intermittently fails because they reach out to an external
+// server. We should replace them with content served locally
 describe('fetch_html', () => {
+    const doctype = '<!DOCTYPE html>';
+
+    it('works', function () {;
```
manuq: This new "works" test works for me. I run the tests with: `./node_modules/.bin/mocha --recursive --check-leaks`.

```
  fetch_html
    ✓ works (1757ms)
```
So I suggest moving the comment about "This test fails" to the "can handle gzipped responses" instead of applying it to the two "fetch html" tests. Or further, either comment or remove the failing test.

manuq: Ah no I was wrong, it is in fact intermittently working/failing! But after looking at the error message `For async tests and hooks, ensure "done()" is called` and looking at mocha docs, I found a way to make these tests work:

```
    it('works', done => {
        const test_url = 'https://creativecommons.org/blog';
        return util.fetch_html(test_url).then(result => {
            expect(result.html().substring(0, doctype.length)).to.equal(doctype);
        }).finally(done());
    });

    it('can handle gzipped responses', done => {
        const test_url = 'https://www.kapanlagi.com/' +
                        'intermezzone/' +
                        'bule-amerika-ini-nyoba-makan-buah-duku-ekspresinya-nggak-nahan-aee243.html';
        return util.fetch_html(test_url).then((result) => {
            expect(result.html().substring(0, doctype.length)).to.equal(doctype);
        }).finally(done());
    });
```

manuq: Also we are favouring arrow functions `=>` instead of `function ()`.

ptomato: Are arrow functions a problem in Mocha tests? In Jasmine tests, which look very similar, I avoid them because Jasmine can use the `this` object passed to the test functions to store state (although that is optional.)

ptomato:
I kept the comment, as I think it is still valid; the test can fail if the external websites are not reachable, and that's bad practice for unit tests.

PR #115, opened 2018-01-30
T20761: Add 'wrapper' and 'wraps' properties to video objects and articles

smspillaz:
This PR adds a 'wrapper' and 'wraps' property to video objects and articles respectively to indicate a "thinly-wraps" relationship. This kind of relationship is currently seen with the video apps that make use of `GalleryVideoArticle` - a piece of HTML that just has a title and embedded video to play in the respective application.

Some consumers of content aren't particularly interested in these wrapper articles and want to just show the underlying content directly. They need some way of excluding the wrapper articles from the search results and including the wrapped content instead. These two properties are indicative of the fact that such a relationship exists between an article and a media object and the shard packers and tag objects accordingly.

https://phabricator.endlessm.com/T20761

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -569,6 +569,7 @@ class GalleryVideoArticle extends BaseAsset {
         Object.assign(metadata, {
             'document': this._document.html(),
             'authors': [this._author],
+            'wraps': [this._main_video],
         });
         return metadata;
     }
--- a/lib/index.js
+++ b/lib/index.js
@@ -252,6 +252,24 @@ class VideoAsset extends BaseAsset {
 
     set_download_uri(value) { this._download_uri = value; }
 
+    // Set a wrapper article asset ID on this video. This way,
+    // consumers can know if this article is merely just wrapping
+    // another video asset and can go directly to the video if
+    // appropriate
+    set_wrapper(article) {
+        // We can't really do validation here, any article could wrap
+        // a video asset and all article types derive from BaseAsset.
+        this._wrapper = article.asset_id;
+    }
+
+    to_metadata() {
+        const metadata = super.to_metadata();
+        Object.assign(metadata, {
+            'wrapper': this._wrapper,
+        });
+        return metadata;
+    }
+
     _save_to_hatch(hatch) {
         hatch._videos.push({
             asset_id: this.asset_id,


smspillaz:
Note: I don't know how to test `libingester` changes so this changeset is effectively untested. However, I have tested the rest of the content build pipeline and it works fine.

lib/index.js
```diff
@@ -551,6 +565,7 @@ class GalleryVideoArticle extends BaseAsset {
         Object.assign(metadata, {
             'document': this._document.html(),
             'authors': [this._author],
+            'wraps': [this._main_video]
```
cosimoc: Nit: add terminal comma

lib/index.js
```diff
@@ -252,6 +252,20 @@ class VideoAsset extends BaseAsset {
 
     set_download_uri(value) { this._download_uri = value; }
 
+    // Set a wrapper article asset ID on this video. This way,
+    // consumers can know if this article is merely just wrapping
+    // another video asset and can go directly to the video if
+    // appropriate
+    set_wrapper(article) { this._wrapper = article.asset_id; }
```
cosimoc: Should there be some enforcement that these are `GalleryVideoArticle` articles? Or that they otherwise have a `wraps` property defined?

smspillaz: There could be. A "wraps" relationship doesnt just have to exist between a GalleryVideoArticle and a VideoAsset, it could exist between any two things. I see your point about restricting it in practice though. Is that your preference?

smspillaz: Okay I did that for now, since I think it makes sense.

lib/index.js
```diff
@@ -252,6 +252,27 @@ class VideoAsset extends BaseAsset {
...
 
+    // Set a wrapper article asset ID on this video. This way,
+    // consumers can know if this article is merely just wrapping
+    // another video asset and can go directly to the video if
+    // appropriate
+    set_wrapper(article) {
+        // The only valid wrapper for VideoAsset is GalleryVideoArticle
+        if (!(article instanceof GalleryVideoArticle)) {
+            throw new Error(`The only valid wrapper for VideoAsset is GalleryVideoArticle`);
+        }
```
cosimoc: Thanks for adding this. Thinking again about this, we could instead have a `wrappable` property that is `false` by default and set to `true` in the `GalleryVideoArticle` subclass

lib/index.js
```diff
@@ -252,6 +252,27 @@ class VideoAsset extends BaseAsset {
...
     set_download_uri(value) { this._download_uri = value; }
 
+    // Set a wrapper article asset ID on this video. This way,
+    // consumers can know if this article is merely just wrapping
+    // another video asset and can go directly to the video if
+    // appropriate
+    set_wrapper(article) {
+        // The only valid wrapper for VideoAsset is GalleryVideoArticle
+        if (!(article instanceof GalleryVideoArticle)) {
+            throw new Error(`The only valid wrapper for VideoAsset is GalleryVideoArticle`);
```
ptomato: This seems incorrect to me. Third parties should be able to write their own article format for their ingester that inherit from `BaseArticle` if they don't want any of the formats that we provide, so I don't think you can make this restriction.

cosimoc: +1, I like this approach as well.

smspillaz: Okay. I think in that case I'll make the `instanceof` check operate on `BaseArticle` for now.

smspillaz:
> Like I said on the issue, I think you should be able to go from the asset to the parent article regardless of whether it is a "thin wrapper" or not. It seems to me that should be able to be done internally in libingester without any API.

At least on the libingester side of things the "thinly wraps" relationship and the "asset is embedded" relationships are two things that need to be kept conceptually separate.

For my usecase, I need to be able to explicitly exclude all articles that are merely "thin wrappers" and explicitly include all assets that are "thinly wrapped". Expanding the definition of "wrapper" to "all embedded assets" would not satisfy my usecase.

Its worth noting that I don't actually need to know *what* the relation is, but rather only *that* an article or asset is a member of that relation.

I think the capturing the broader relation of "who embeds this asset" is useful, but probably something to be done in a separate story to avoid confusion (its also worth noting that capturing that relation only works if articles and assets are in a one-to-many relationship, something that may not hold in the future if we ever want to deduplicate assets).

lib/index.js
smspillaz: Damn, we can't check for a base class here either - the base of all articles is `BaseAsset` - there is no intermediate class that we can check for.

smspillaz:
Closing for now, we'll want to remove the wrapper articles entirely in an API bump.

PR #116, opened 2018-01-30
previewer: Improve sidebar usability

manuq:
As a regular user of libingester-previewer, I find a bit confusing to
see the selected row in the sidebar, because of the striped rows.  It
also makes the text bold in the selected row, resizing the text and
flickering because often the line number changes.

This commit removes striped rows, and uses the bootstrap class to make
rows active.


sgnn7:
CR Pass

PR #117, opened 2018-01-30
Fixes for iconv

ptomato:
We don't need to use iconv at all, as long as we stick to NodeJS's built-in encodings.

https://phabricator.endlessm.com/T20985


sgnn7:
Great detective work! CR Pass!

PR #118, opened 2018-01-31
API docs!

ptomato:
https://phabricator.endlessm.com/T21011


lib/index.js
```diff
@@ -627,25 +781,133 @@ class NewsArticle extends BaseAsset {
...
         this._main_image = asset;
         this._main_image_caption = _ensure_cheerio(caption);
     }
 
     set_tags(value) { throw new Error(`Use set_section() for NewsArticle`); }
+    /**
+     * Pass in the ID string of a "set" (category, tag). This is not the
+     * human-readable string (though it's possible for the two to be
+     * identical); the human-readable name of the set is created later in your
+     * permanent shard.
```
cosimoc: libingester really does not know/care about shards, so it feels a bit weird to see it mentioned here.

lib/index.js
```diff
@@ -627,25 +781,133 @@ class NewsArticle extends BaseAsset {
...
+     *
+     * (Compare {@link BlogArticle#set_tags} for the blog article format; news
+     * articles are usually in one section of the news site at a time. If it
+     * turns out we need more tags, we can also add a `set_tags()` method to
+     * the news article format.)
```
cosimoc: This last sentence looks like something we should put in a `TODO` comment in the code, rather in the documentation block.

lib/index.js
```diff
@@ -88,23 +88,72 @@ function _ensure_synopsis (value) {
...
+    /**
+     * @param {number} value
+     */
     set_sequence_number(value) { this._sequence_number = value; }
+    /**
+     * @param {string} value
+     */
     set_license(value) { this._license = value; }
+    /**
+     *  Pass in an array of ID strings for "sets" (categories, tags). The ID
```
cosimoc: Extra leading whitespace

lib/index.js
```diff
@@ -88,23 +88,72 @@ function _ensure_synopsis (value) {
...
+     * string is not the human-readable string (though it's possible for the
+     * two to be identical); the human-readable name of the set is created
+     * later in the app builder.
```
cosimoc: I would not mention the app builder here in the documentation block, since it's not something that's necessarily accessible to the ingester writer.

docs/assets/site.js
```diff
@@ -0,0 +1,168 @@
...
+}
+
+function showHashTarget(targetId) {
+  if (targetId) {
+    var hashTarget = document.getElementById(targetId);
+    // new target is hidden
+    if (
+      hashTarget &&
+      hashTarget.offsetHeight === 0 &&
+      hashTarget.parentNode.parentNode.classList.contains('display-none')
```
sgnn7: I don't think JS complains if you try to remove the class that's isn't there so this last case can be removed.

docs/index.html
```diff
@@ -0,0 +1,3525 @@
+<!doctype html>
```
sgnn7: Whatever tool we're using is creating some really bad HTML styling (bad indentation, bad nesting, extraneous newlines, etc). Not sure if we should try to fix this in the tooling.

lib/index.js
ptomato: "Set" is the preferred term in the modular framework's API, it was a term picked for that purpose by the designers. I think it would be best to keep it.

lib/index.js
ptomato: Oops! Copy-pasta error from our internal wiki. I changed this to shards (see comment below).

lib/index.js
ptomato: I guess not, but how else should we answer the question "Where do I create the human-readable name?"

docs/assets/anchor.js
```diff
@@ -0,0 +1,350 @@
+/*!
```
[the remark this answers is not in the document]
ptomato: I would suggest not reviewing the `docs` directory, it's entirely autogenerated by the documentation tool.

lib/index.js
```diff
@@ -88,23 +88,72 @@ function _ensure_synopsis (value) {
...
+/** */
 class BaseAsset {
     constructor() {
         this._asset_id = new_asset_id();
         this._ekn_tags = [];
         this._tags = [];
     }
 
+    /**
+     * 20-byte hex ID of the document
```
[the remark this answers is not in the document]
ptomato: If it's not a SHA1 hash then it won't be accepted by knowledge-lib! Is there any code that puts UUIDs in here?

lib/index.js
```diff
@@ -627,25 +781,133 @@ class NewsArticle extends BaseAsset {
...
     set_as_static_page() { this._ekn_tags.push(EKN_STATIC_TAG); }
+    /**
+     * Name of the publication, e.g. "New York Times". This shows up in a link
+     * in the top right hand corner of rendered news articles.
+     * @param {string} value - The name of the source
+     */
     set_source(value) { this._source = value; }
+    /**
+     * Sets the lede (opening paragraph) of the news article. This paragraph is
+     * marked up specially.
```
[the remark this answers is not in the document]
ptomato: No, that's the "synopsis", although ingesters often feed the same text (in plaintext, not HTML) to the synopsis.

lib/index.js
sgnn7: The modular framework only makes sense in context of everything else but it's not applicable to someone just writing a simple ingester and using the previewer to check their work thus the comment. I don't feel that strongly about it so it's fine if left as-is.

docs/assets/anchor.js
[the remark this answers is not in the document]
sgnn7: @ptomato Oh - I thought that just the index.html was the autogenerated code. If that's not the case, ignore the comments on those files.

lib/index.js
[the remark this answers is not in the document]
sgnn7: Hmm ... maybe I'm mistaken here then... @wgreenberg should know though.

lib/index.js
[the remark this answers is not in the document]
sgnn7: Fair enough.

sgnn7:
CR Pass

PR #125, opened 2018-02-08
T20761: Add VideoArticle

smspillaz:
This class is just a wrapper on top of `VideoAsset` which retains the "authors" metadata. It should be preferred over `GalleryVideoArticle` because it doesn't interpose a wrapper page, which was proving problematic for the apps.

https://phabricator.endlessm.com/T20761

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -695,6 +695,8 @@ class GalleryVideoArticle extends BaseAsset {
         super();
         this._object_type = "ArticleObject";
         this._content_type = "text/html";
+
+        console.warn('GalleryVideoArticle is deprecated. Use VideoArticle instead.');
     }
 
     set_author(value) { this._author = value; }
--- a/lib/index.js
+++ b/lib/index.js
@@ -781,6 +781,43 @@ class GalleryVideoArticle extends BaseAsset {
 
 exports.GalleryVideoArticle = GalleryVideoArticle;
 
+/**
+ * A simple wrapper over VideoAsset that should be used for toplevel videos.
+ *
+ * This asset type is for videos that aren't embedded as part of another
+ * piece of content, but should instead be directly accessible from
+ * content listings and search results. For instance, if you were building
+ * an app where the only content was videos and those videos required no
+ * other context other than some displayable metadata such as the
+ * authors and a brief synopsis, then you should use this asset type.
+ *
+ * In the past, developers used GalleryVideoArticle, which wrapped
+ * the video in another HTML page. That asset type is now deprecated
+ * since apps are capable of displaying the video and its corresponding
+ * metadata correctly without the need for a page in between.
+ */
+class VideoArticle extends VideoAsset {
+
+    /**
+     * Set the author of this document to the given value.
+     * The author should be the "creator" of the video, being
+     * the person or organization responsible for its production.
+     *
+     * @param {string} value - The name of the author.
+     */
+    set_author(value) { this._author = value; }
+
+    to_metadata() {
+        const metadata = super.to_metadata();
+        Object.assign(metadata, {
+            'authors': [this._author]
+        });
+        return metadata;
+    }
+}
+
+exports.VideoArticle = VideoArticle;
+
 /** */
 class NewsArticle extends BaseAsset {
     constructor() {
--- a/docs/index.html
+++ b/docs/index.html
@@ -257,6 +257,34 @@
               
                 
                 <li><a
+                  href='#videoarticle'
+                  class=" toggle-sibling">
+                  VideoArticle
+                  <span class='icon'>▸</span>
+                </a>
+                
+                <div class='toggle-target display-none'>
+                  
+                  
+                    <ul class='list-reset py1-ul pl1'>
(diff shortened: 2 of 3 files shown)


lib/index.js
```diff
@@ -687,6 +687,8 @@ class GalleryVideoArticle extends BaseAsset {
         super();
         this._object_type = "ArticleObject";
         this._content_type = "text/html";
+
+        console.warn('GalleryVideoArticle is deprecated. Use VideoArticle instead.');
```
ptomato: Do we unconditionally want to switch to `VideoArticle` in all cases?

lib/index.js
```diff
@@ -771,6 +773,21 @@ class GalleryVideoArticle extends BaseAsset {
 
 exports.GalleryVideoArticle = GalleryVideoArticle;
 
+/** */
```
ptomato: Do we want to make it public API yet? If not, best to leave this out so it doesn't show up in the API reference. If yes, then document it :smile: (the `set_author()` method would need documentation as well)

If you document it, you'll also need to add a commit regenerating the documentation. Use `npm run docs:publish` before pushing your branch.

lib/index.js
smspillaz: I'm not sure about doing that just yet - I'd prefer to test each one rather than forcing a switch.

lib/index.js
smspillaz: Okay, will do

smspillaz: Done

lib/index.js
ptomato: I meant the other way around — we don't necessarily want a deprecation message if there are still legitimate uses of `GalleryVideoArticle`.

lib/index.js
```diff
@@ -771,6 +773,31 @@ class GalleryVideoArticle extends BaseAsset {
 
 exports.GalleryVideoArticle = GalleryVideoArticle;
 
+/**
+ * A simple wrapper over VideoAsset that should be used for toplevel videos.
+ *
+ * Along with the usual VideoAsset methods, it also supports set_author,
+ * which allows it to be used in place of GalleryVideoArticle.
```
ptomato: editorial nitpick: We never actually exposed GalleryVideoArticle in the API docs, so best not to mention it here. Try to give a bit more context about what this type of asset is for. For example, something like "This kind of asset is used for videos that aren't embedded as part of other content, where the video itself is the main item."

lib/index.js
```diff
@@ -771,6 +773,31 @@ class GalleryVideoArticle extends BaseAsset {
...
+ */
+class VideoArticle extends VideoAsset {
+
+    /**
+     * Set the author of this document to the given value.
```
ptomato: More editorial comments: it would be a bonus to mention how "author" relates to video, i.e. that the author is the creator of the video.

lib/index.js
smspillaz: Hmm. I would like to warn future users against using it, since generally speaking, I think we want to migrate away from it in all cases. The concern I had was about migrating away from it piecemeal as opposed to forcing a potentially breaking change.

lib/index.js
smspillaz: Yup, fair enough.

lib/index.js
smspillaz: I think I would prefer to keep the mention here so that the migration path is made clear. But I'll add the context.

lib/index.js
```diff
@@ -781,6 +781,43 @@ class GalleryVideoArticle extends BaseAsset {
...
 
+/**
+ * A simple wrapper over VideoAsset that should be used for toplevel videos.
+ *
+ * This asset type is for videos that aren't embedded as part of another
+ * piece of content, but should instead be directly accessible from
+ * content listings and search results. For instance, if you were building
+ * an app where the only content was videos and those videos required no
+ * other context other than some displayable metadata such as the
+ * authors and a brief synopsis, then you should prefer to use this asset type.
```
cosimoc: Nitpick: `prefer to use this asset type`. Prefer over what? Perhaps just remove `prefer to` and just saw `you should use`.

lib/index.js
```diff
@@ -781,6 +781,43 @@ class GalleryVideoArticle extends BaseAsset {
...
+ *
+ * In the past, developers used GalleryVideoArticle, which wrapped
+ * the video in another HTML page. That asset type is now deprecated
+ * since apps are capable of displaying the video and their corresponding
```
cosimoc: `displaying the video and their` -> `displaying videos and their` or `displaying the video and its`

lib/index.js
```diff
@@ -781,6 +781,43 @@ class GalleryVideoArticle extends BaseAsset {
...
+ * metadata correctly without the need for a page in between.
+ */
+class VideoArticle extends VideoAsset {
+
+    /**
+     * Set the author of this document to the given value.
+     * The author should be the "creator" of the video, being
+     * the person or organisation responsible for its production.
```
cosimoc: `organisation` -> `organization` -- our docs are usually in `en_US` :)

PR #126, opened 2018-02-09
Document more

ptomato:
Documents some more API entry points that I'm using in the ingester walkthrough.

https://phabricator.endlessm.com/T20978


sgnn7:
CR Pass

PR #127, opened 2018-02-09
lib: Add set_parent_asset_id to BaseAsset

smspillaz:
This can be used to explicitly indicate which asset ID is the
"parent" to this one, for instance, the asset which embeds a video
or an image.

It needs to be set explicitly, since we don't have a guarantee that
toplevel assets will be processed and saved before child assets, so
there's no way to look up an asset's parents at processing time.

https://phabricator.endlessm.com/T21114

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -98,6 +98,12 @@
                       </a></li>
                       
                       <li><a
+                        href='#baseassetset_parent'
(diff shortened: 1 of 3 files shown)


smspillaz:
(Not the approach we wanted)

PR #130, opened 2018-02-28
T21345/reingest

manuq:
https://phabricator.endlessm.com/T21345

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -6,11 +6,21 @@ const crypto = require('crypto');
 const mustache = require('mustache');
 const path = require('path');
 const sass = require('node-sass');
+const argv = require('argv-parse');
 const verify = require('./verify');
 
 const thenify_all = require('thenify-all');
 const fs = thenify_all(require('fs-extra'), {}, ['readFile', 'writeFile', 'readdir', 'copy']);
 const tar = require('tar');
+
+const ARGV_OPTIONS = {
+    'no-tgz': {
+        type: 'boolean'
+    },
+    'path': {
+        type: 'string'
+    }
+};
 
 const EKN_STATIC_TAG = 'EknStaticTag';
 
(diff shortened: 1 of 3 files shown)


sgnn7:
LGTM but I'll leave the full CR and merge to @dnohales

lib/index.js
```diff
@@ -1097,23 +1131,19 @@ class Hatch {
...
-            const no_tgz = argv.indexOf('--no-tgz');
-            if (no_tgz >= 0) {
-                options.no_tgz = true;
-            }
+        const argv_options = options.argv || process.argv.slice(2);
+
+        const config_options = argv(ARGV_CONFIGFILE_OPTIONS, argv_options);
+        if (config_options['config-file']) {
+            const cf = config_options['config-file'];
+            const file_options = JSON.parse(fs.readFileSync(cf, 'utf8'));
```
dnohales: The package `fs-extra` has a function `readJsonSync`.

lib/index.js
```diff
@@ -1097,23 +1131,19 @@ class Hatch {
...
+            options = Object.assign(options, file_options);
         }
 
-        return options;
+        return Object.assign(options, argv(ARGV_OPTIONS, argv_options))
```
dnohales: You could easily have only one `ARGV_OPTIONS` variable with `config-file` in it, use it in line 1139, and here do:

```
return Object.assign(file_options, config_options, options);
```

(arguments on the right have priority)

lib/index.js
```diff
@@ -1097,23 +1131,19 @@ class Hatch {
...
+    _parse_options(options) {
+        options = options || {};
 
-            const no_tgz = argv.indexOf('--no-tgz');
-            if (no_tgz >= 0) {
-                options.no_tgz = true;
-            }
+        const argv_options = options.argv || process.argv.slice(2);
+
+        const config_options = argv(ARGV_CONFIGFILE_OPTIONS, argv_options);
```
dnohales: I'd name this `cli_options`

lib/index.js
```diff
@@ -6,12 +6,31 @@ const crypto = require('crypto');
...
+};
+
+const ARGV_OPTIONS = {
+    'no-tgz': {
+        type: 'boolean'
+    },
+    'path': {
+        type: 'string'
+    },
+    'urls': {
```
dnohales: Maybe `reingest-urls`? Not a blocker though.

manuq: "reingest" is a good verb for SOMA, but the purpose of 'urls' can be any, like ingesting a specific URL for debugging.

PR #131, opened 2018-03-09
Document hacking

manuq:


HACKING.md
```diff
@@ -0,0 +1,8 @@
+# Release checklist #
+
+- [ ] Run `npm run docs:publish`. This will create an automatic commit
+- [ ] Run `npm version` to increment the version according to semver
+      guidelines. For example: `npm version patch`. Check `npm version
+      --help` for other options.
+- [ ] Push.
+- [ ] Run `npm publish`.
```
sgnn7: Hmm.... Can we automate all of this? A script that must be used for bumping the version that regens docs in the process should ensure we never have stale docs.

manuq:
Indeed, automating the release is much better. I added a commit with a script to automate it. Do you think we need any other checks, besides the branch == master?

run_release.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -e
```
sgnn7: you can tack this onto the shebang line `#!/bin/bash -e`

run_release.sh
```diff
@@ -0,0 +1,18 @@
...
+
+npm_version_params=$*
+
+current_branch=$(git rev-parse --abbrev-ref HEAD)
```
sgnn7: I would love to see some sort of a yes/no question before doing the rest of the code to prevent accidental changes. Not a blocker but it would be neat to make a git tag for that version too maybe.

run_release.sh
```diff
@@ -0,0 +1,18 @@
...
+if [[ "$current_branch" != "master" ]]; then
+    echo 'Not in git master. Aborting release';
+    exit 1;
+fi
+
+new_tag=$(npm version $npm_version_params)
+npm run docs:publish
+git push
```
sgnn7: Why the unqualified git push?

sgnn7: Hmm..maybe we should commit to repo too.

run_release.sh
manuq: +1. The `npm version` below automatically creates a new tag.

run_release.sh
manuq: Should this be `git push origin master`?

sgnn7: master + tags should only need one (`git push origin --tags master`)

sgnn7:
LGTM as well. I guess it's merge time?

PR #133, opened 2018-03-15
new API get_max_settings()

manuq:
To reuse the logic of util.fetch_rss_entries() in the ingesters, and
do the environment variable parsing in one place.

https://phabricator.endlessm.com/T20341

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -6,7 +6,6 @@ const cheerio = require('cheerio');
 const crypto = require('crypto');
 const feedparser = require('feedparser');
 const libingester = require('./index');
-const moment = require('moment');
 const parseDataUrl = require('parse-data-url');
 const request = require('request');
 const somaDOM = require('./soma_dom');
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -1385,4 +1385,34 @@ class Hatch {
...
+ *
+ * The `max_items` and `max_days_old` parameters can be overridden by the
+ * `LIBINGESTER_MAX_ITEMS` and `LIBINGESTER_MAX_DAYS_OLD` environment
+ * variables, respectively.
+ *
+ * @param {number} [max_items=Infinity] - Maximum number of entries to ingest
+ * @param {number} [max_days_old=1] - Maximum age of entries to ingest, in days
+ * @returns {Object} - An object with parsed settings `max_items` and `oldest_date`
+ */
+function get_max_settings(max_items=Infinity, max_days_old=1) {
```
cosimoc: `get_max_settings()` is a pretty awkward name... I would rename this to e.g. `get_ingestion_settings()`, or `get_ingestion_limits()`. Also it may be worth moving it to the util section.

lib/util.js
```diff
@@ -314,20 +314,10 @@ exports.encode_uri = encode_uri;
...
-            max_items = parseInt(process.env.LIBINGESTER_MAX_ITEMS);
-        }
-        if (process.env.LIBINGESTER_MAX_DAYS_OLD) {
-            max_days_old = parseInt(process.env.LIBINGESTER_MAX_DAYS_OLD);
-        }
-    } catch (err) {
-        throw new Error(`Couldn't parse RSS entry overrides: ${err}`);
-    }
-
-    const oldest_date = moment().subtract(max_days_old, 'days');
```
cosimoc: Old code used the external `moment` library to achieve this. I am happy to change it, but we should also remove the dependency and the require in that case.

manuq: Good catch, I wanted to remove moment too and forgot about it in the last minute.

PR #134, opened 2018-03-15
util: make wordpress paginator also work with an array

cosimoc:
https://phabricator.endlessm.com/T21595


lib/util.js
```diff
@@ -387,13 +387,24 @@ function _fetch_rss_page (feed, page, items, all_items, max_items, oldest_date)
 
 exports.fetch_rss_entries = fetch_rss_entries;
 
+function create_wordpress_paginator (feed) {
+    return function (n) {
```
sgnn7: Might be easier to read with `n` expanded to a full var name

cosimoc: Fixed

sgnn7:
Should I merge this or leave it to @manuq to look over too?

cosimoc:
The review request was for either of you :-)
Let's get it in!

PR #135, opened 2018-03-16
Improve util.get_embedded_video_asset()

ptomato:
https://phabricator.endlessm.com/T21578


manuq:
Yeah, making the DOM element with Cheerio is much better.

PR #136, opened 2018-03-17
Save video metadata to the hatch as well

ptomato:
Previously, any metadata set on a VideoAsset was discarded.

https://phabricator.endlessm.com/T21578

### DIFF
--- a/lib/verify.js
+++ b/lib/verify.js
@@ -18,7 +18,6 @@ function verify_metadata(metadata) {
 
     assert((metadata['assetID'].length === 40), "Asset has invalid assetID");
     assert(typeof metadata['canonicalURI'] === "string", "Asset has invalid canonicalURI");
-    assert(typeof metadata['contentType'] === "string", "Asset has invalid contentType");
 
     assert(!!metadata['matchingLinks'], "Asset missing matchingLinks");
     assert(metadata['matchingLinks'].every((s) => (typeof s === "string")), "Some asset matching URIs are not strings");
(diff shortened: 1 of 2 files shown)


lib/verify.js
```diff
@@ -41,6 +43,8 @@ function verify_metadata(metadata) {
     } else if (object_type === "DictionaryWordObject") {
         assert(!!metadata['word'], "metadata missing word");
         assert(!!metadata['definition'], "metadata missing definition");
+    } else if (object_type === 'VideoObject') {
+        assert(!metadata['contentType'], "Video object should have its contentType set later");
```
sgnn7: It's possible that the postal uses this info. @milossymphony do you know if that's the case? I can check too if you don't know.

milossymphony: @sgnn7 Not sure I understand the question.  But there is this piece of code when mapping hatch manifest items to content items:
```
  // Decorate the manifest items with some metadata so we can distinguish
  // between them later on
  manifest.assets.forEach(item => {
    item.isHatchAsset = true;
    item.contentType = 'application/x-hatch-asset';
  });

  // This isn't an actual mimetype, but that's fine since our ingestion
  // process is looking for job-system specific values here
  manifest.videos.forEach(item => {
    item.contentType = 'Video';
  });
```

sgnn7: Ah, ok - then we're good. I didn't know if the mapping was based on mimeType but it looks like it's based on manifest asset type so I think we're good.

sgnn7:
CR Pass

PR #138, opened 2018-03-20
Wip rss pagination

manuq:
https://phabricator.endlessm.com/T21642


lib/util.js
```diff
@@ -373,20 +373,37 @@ function _fetch_rss_json (feed_url) {
...
+          feed_json.meta['rss:site']['@']['xmlns'].startsWith('com-wordpress');
+    return has_generator || has_xmlns;
+}
+
 function _fetch_rss_page (feed, page, items, all_items, max_items, oldest_date) {
-    let feed_obj, is_paginated;
-    if (typeof feed === 'function') {
-        is_paginated = true;
-        feed_obj = feed(page);
+    let is_array = false;
```
sgnn7: Maybe overly complex? You could do `is_array = Array.isArray(feed)` even though you do another check later.

lib/util.js
```diff
@@ -373,20 +373,37 @@ function _fetch_rss_json (feed_url) {
...
-    let feed_url = feed_obj, is_array = false;
-    if (Array.isArray(feed_obj)) {
-        is_array = true;
-        feed_url = feed_obj.shift();
+    if (typeof feed_obj === 'function') {
+        is_paginated = true;
+        feed_url = feed_obj(page);
+    } else {
+        feed_url = feed_obj;
+    }
```
sgnn7: You could remove the `else` if you set feed_url to feed_obj before the if (reduces nesting)

lib/util.js
```diff
@@ -398,31 +415,54 @@ function _fetch_rss_page (feed, page, items, all_items, max_items, oldest_date)
...
-            if (is_paginated)
+        if ((typeof feed_obj !== 'function') && _is_wordpress_feed(feed_json)) {
+            is_paginated = true;
+            const new_feed = _do_create_wordpress(feed_obj);
+            if (is_array) {
+                feed.shift();
+                feed = [new_feed].concat(feed);
+            } else {
+                feed = new_feed;
+            }
```
sgnn7: Same thing - you could simplify to remove the `else`

lib/util.js
```diff
@@ -398,31 +415,54 @@ function _fetch_rss_page (feed, page, items, all_items, max_items, oldest_date)
...
-            return `${feed}?paged=${page_num}`;
-        }
+        return `${feed}?paged=${page_num}`;
     };
 }
 
+function create_wordpress_paginator (feed) {
+    if (Array.isArray(feed)) {
+        return feed.map(uri => _do_create_wordpress(uri));
+    } else {
```
sgnn7: There's no need for else - you only get below this line if it's not an array

lib/util.js
```diff
@@ -373,20 +373,37 @@ function _fetch_rss_json (feed_url) {
...
+}
+
 function _fetch_rss_page (feed, page, items, all_items, max_items, oldest_date) {
-    let feed_obj, is_paginated;
-    if (typeof feed === 'function') {
-        is_paginated = true;
-        feed_obj = feed(page);
+    let is_array = false;
+    let is_paginated = false;
+    let feed_url, feed_obj;
```
sgnn7: Hmm... we need to figure out if multiple var declarations on a single line is something we want to do but since we don't have rules for it yet, not a blocker.

ptomato: Pick one and add it to eslint :-)

lib/util.js
```diff
@@ -398,31 +415,54 @@ function _fetch_rss_page (feed, page, items, all_items, max_items, oldest_date)
...
+            } else {
+                feed = new_feed;
+            }
+        }
+
+        if (done_crawling) {
+            if (is_paginated && is_array) {
+                feed.shift();
+                return _fetch_rss_page(feed, 1, new_items, new_all_items, max_items, oldest_date);
+            }
```
cosimoc: I don't understand this code... If we're done crawling we should not fetch anything else. It may be worth moving the check above, before creating `new_all_items`

lib/util.js
```diff
@@ -398,31 +415,54 @@ function _fetch_rss_page (feed, page, items, all_items, max_items, oldest_date)
...
         const done_crawling = (limited_items.length < feed_json.items.length);
 
-        if ((is_paginated || is_array) && !done_crawling) {
-            const new_all_items = new Set(function* () { yield* all_items; yield* new_items.map(item => item.link); }());
+        const new_all_items = new Set(function* () { yield* all_items; yield* new_items.map(item => item.link); }());
 
-            if (is_paginated)
+        if ((typeof feed_obj !== 'function') && _is_wordpress_feed(feed_json)) {
+            is_paginated = true;
+            const new_feed = _do_create_wordpress(feed_obj);
```
cosimoc: Can't we use `create_wordpress_paginator()` here directly?

lib/util.js
sgnn7: We use jshint and it looks like there's no support for this (even though I found some info talking about `mutltivar` which turns out it doesn't work). Maybe we should convert to eslint linting?

cosimoc:
This could probably also use a test case.

sgnn7:
@manuq needs a rebase

lib/util.js
```diff
@@ -398,31 +412,57 @@ function _fetch_rss_page (feed, page, items, all_items, max_items, oldest_date)
...
+
+        const _continue = (done_crawling && is_paginated && Array.isArray(feed)) ||
+              (!done_crawling && (is_paginated || Array.isArray(feed)));
+
+        if (!_continue) {
             return new_items;
         }
+
+        let new_page;
+        if (done_crawling && is_paginated && Array.isArray(feed)) {
```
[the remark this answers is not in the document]
manuq: @cosimoc yes, "done crawling" in the case of an array means done for the current RSS item. To illustrate the bug, run this script in master enabling debug logs like this: `DEBUG=libingester LIBINGESTER_MAX_DAYS_OLD=60 node index.js`
```
const libingester = require('libingester');
const logger = libingester.logging.logger;

RSS_FEED = ['https://creativecommons.org/blog/feed/', 'https://girleatworld.net/feed/'];
RSS_FEED = libingester.util.create_wordpress_paginator(RSS_FEED);

libingester.util.fetch_rss_entries(RSS_FEED)
    .then(items => {
        logger.info('how many items?', items.length);
    });
```
With current master, only the first RSS is paginated:

```
$ DEBUG=libingester LIBINGESTER_MAX_DAYS_OLD=60 node index.js
debug: fetching RSS https://creativecommons.org/blog/feed/?paged=1
debug: fetching RSS https://creativecommons.org/blog/feed/?paged=2
info: how many items? 12
```

With this branch, the second RSS is fetched, as it should:

```
$ DEBUG=libingester LIBINGESTER_MAX_DAYS_OLD=60 node index.js
debug: fetching RSS https://creativecommons.org/blog/feed/?paged=1
debug: fetching RSS https://creativecommons.org/blog/feed/?paged=2
debug: fetching RSS https://girleatworld.net/feed/?paged=1
info: how many items? 16
```

Also with this branch, you can remove the line `create_wordpress_paginator`, and still works!

```
$ DEBUG=libingester LIBINGESTER_MAX_DAYS_OLD=60 node index.js
debug: fetching RSS https://creativecommons.org/blog/feed/
debug: Wordpress feed found, adding pagination
debug: fetching RSS https://creativecommons.org/blog/feed/?paged=2
debug: fetching RSS https://girleatworld.net/feed/
debug: Wordpress feed found, adding pagination
info: how many items? 16
```

lib/util.js
```diff
@@ -424,21 +438,51 @@ function _fetch_rss_page (feed, page, items, all_items, max_items, oldest_date)
         // number of items, cease crawling
         const done_crawling = (limited_items.length < feed_json.items.length);
 
-        if ((is_paginated || is_array) && !done_crawling) {
-            const new_all_items = new Set(function* () { yield* all_items; yield* new_items.map(item => item.link); }());
+        const new_all_items = new Set(function* () { yield* all_items; yield* new_items.map(item => item.link); }());
```
cosimoc: I would move this down closer to where it's used, since it's not used in every code branch here.

lib/util.js
```diff
@@ -424,21 +438,51 @@ function _fetch_rss_page (feed, page, items, all_items, max_items, oldest_date)
...
+
+        if ((typeof feed_obj !== 'function') && _is_wordpress_feed(feed_json)) {
+            logger.debug('Wordpress feed found, adding pagination');
+            is_paginated = true;
+            const new_feed = _do_create_wordpress(feed_obj);
+            if (Array.isArray(feed)) {
```
cosimoc: We call `isArray()` many times in this function; I would prefer if we factored that out in a separate variable.

lib/util.js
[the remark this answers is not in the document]
cosimoc: OK, that makes sense! Can we perhaps rename the `done_crawling` variable to another name?

manuq:
I added more commits with test coverage, but the WIP branch needs more work, so I'm closing the PR for now.

PR #139, opened 2018-03-21
Add logging using winston library

manuq:
- Move all console.log to the corresponding winston: info, debug,
  warn, error

- Add debug lines for HTML and RSS fetching.  With this we can remove
  console logs from ingesters.

https://phabricator.endlessm.com/T21703


lib/index.js
```diff
@@ -1334,10 +1349,10 @@ class Hatch {
             return this._create_hatch_archive();
         })
         .then((tarballPath) => {
-            console.log(`hatch: finished, saved to ${tarballPath}`);
+            winston.info(`hatch: finished, saved to ${tarballPath}`);
         })
         .catch((err) => {
```
ptomato: Not necessarily in scope for this pull request, but is this catch clause redundant now?

manuq: True, we configure winston to handle exceptions. I removed it in a separate commit.

manuq:
@ptomato I modified the commits. It turns out the catch() is necessary because winston doesn't log unhandled promise rejections (a promise is automatically rejected wihen a exception happens in it). But, we can add a callback to the `unhandledRejection` event, then log and exit with return code 1. With this we can remove this boilerplate at the end of all current ingesters:

```
        .catch(err => {
            console.log(err);
            process.exitCode = 1;
        });
```

Any ingester can still catch() and do whatever it wants.

lib/index.js
```diff
@@ -8,11 +8,17 @@ const path = require('path');
 const sass = require('node-sass');
 const argv = require('argv-parse');
 const verify = require('./verify');
+const {winston} = require('./logging');
```
sgnn7: If you have a logger abstract module, I think we should export `logger` rather than the implementation-specific `winston`

lib/index.js
```diff
@@ -8,11 +8,17 @@ const path = require('path');
...
 
 const thenify_all = require('thenify-all');
 const fs = thenify_all(require('fs-extra'), {}, ['readFile', 'writeFile', 'readdir', 'copy']);
 const tar = require('tar');
 
+process.on('unhandledRejection', ((reason, promise) => {
+    winston.error("Unhandled promise rejection. Reason: ", reason);
+    process.exitCode = 1;
+}));
```
sgnn7: If you're putting a hook on the process, you can do that in the logging module to keep things self-contained

lib/index.js
manuq: Yeah, good point. I will call it `logger`.

lib/index.js
sgnn7: Followup: you can do process.once() maybe? Not sure how it will deal with multiple logging module imports.

manuq: Sounds good. I am also working on better debugging of assets that don't pass the validations. Currently we get a obscure validation error, and is hard to know what caused it.

ptomato:
Let me know when this is deployed, I'll make a change to the ingester walkthrough to remove the catches there.

sgnn7:
@ptomato done (v2.5.6)

PR #140, opened 2018-03-23
Video metadata: fallbacks for canonical URI and matching links

manuq:
When metadata was added for videos in a1d42969, the canonical URI and
matching links became mandatory.  This fallbacks both to the download
URI.

https://phabricator.endlessm.com/T21734

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -361,6 +361,15 @@ class VideoAsset extends BaseAsset {
             hatch._failed_assets.push(this);
         }));
     }
+
+    to_metadata() {
+        const metadata = super.to_metadata();
+        metadata.canonicalURI = metadata.canonicalURI || this._download_uri;
+        if (!metadata.matchingLinks.length || !metadata.matchingLinks[0]) {
+            metadata.matchingLinks = [this._download_uri];
+        }
+        return metadata;
+    }
 }
 
 exports.VideoAsset = VideoAsset;


lib/index.js
```diff
@@ -361,6 +361,18 @@ class VideoAsset extends BaseAsset {
             hatch._failed_assets.push(this);
         }));
     }
+
+    to_metadata() {
+        const metadata = super.to_metadata();
+        if (!metadata.canonicalURI) {
+            metadata.canonicalURI = this._download_uri;
```
sgnn7: you can inline this with an `||` to skip the `if`
```
metadata.canonicalURI = metadata.canonicalURI || this._download_uri;
```

lib/index.js
```diff
@@ -361,6 +361,18 @@ class VideoAsset extends BaseAsset {
...
+        }
+        if (!metadata.matchingLinks.length || !metadata.matchingLinks[0]) {
```
sgnn7: Why do you need to check if the first item is falsey? Will that ever be the case?

lib/index.js
```diff
@@ -361,6 +361,18 @@ class VideoAsset extends BaseAsset {
...
+            metadata.matchingLinks = [this._download_uri];
+        }
+        return metadata;
+    }
+
```
sgnn7: Extra newline not needed

lib/index.js
manuq: yes, see the BaseAsset to_metadata(), it sets `"matchingLinks": [ this._canonical_uri ],` which for the video can be `[ undefined ] `.

sgnn7:
CR Pass

PR #142, opened 2018-04-07
util: Atomatically paginate Wordpress feeds

manuq:
Also fix bugs in pagination: arrays respecting MAX_DAYS_OLD only for
the first item.  Added unit tests.  Also added a number of dev
dependencies to stub the test feeds:

- sinon and rewire in combination to stub the _fetch_rss_json()
  private function.

- sinon.useFakeTimers() to stub the global Date object.

- json-date-parser to obtain Date objects when parsing the json file.

https://phabricator.endlessm.com/T21642

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -398,6 +398,16 @@ function _fetch_rss_json (feed_url) {
     });
 }
 
+function _is_wordpress_feed(feed_json) {
+    const has_generator = feed_json.meta.generator &&
+          feed_json.meta.generator.startsWith('https://wordpress.org/');
+    const has_xmlns = feed_json.meta['rss:site'] &&
+          feed_json.meta['rss:site']['@'] &&
+          feed_json.meta['rss:site']['@']['xmlns'] &&
+          feed_json.meta['rss:site']['@']['xmlns'].startsWith('com-wordpress');
+    return has_generator || has_xmlns;
+}
+
 function _fetch_rss_json_retrying(feed_url, attempt=1) {
     return _fetch_rss_json(feed_url).catch(err => {
         if (attempt > MAX_RETRIES) {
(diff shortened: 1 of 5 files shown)


test/lib/util.test.js
```diff
@@ -1,11 +1,36 @@
...
-const util = require('../../lib/util');
+const rewire = require('rewire');
+const util = rewire('../../lib/util');
 const libingester = require('../../lib/index');
+const { jsonDateParser } = require('json-date-parser');
+
+const FAKE_NOW = 1523065866010;  // 2018-04-07T01:51:06.010Z
+
+const _RSS_JSON = fs.readFileSync(__dirname + '/test_files/rss-feeds.json',
+                                  jsonDateParser);
```
cosimoc: Does passing the parser do anything here? I thought it only did something when passed to `JSON.parse()`

lib/util.js
```diff
@@ -443,23 +456,55 @@ function _fetch_rss_page (feed, page, items, all_items, max_items, oldest_date)
...
+            } else {
+                feed = new_feed;
+            }
+        }
 
-        if ((is_paginated || is_array) && !done_crawling) {
-            const new_all_items = new Set(function* () { yield* all_items; yield* new_items.map(item => item.link); }());
+        const _continue = (done_crawling && is_paginated && Array.isArray(feed)) ||
+              (!done_crawling && (is_paginated || Array.isArray(feed))) ||
+              (done_crawling && !is_paginated && Array.isArray(feed));
```
cosimoc: The first and third clauses are the same, except that `is_paginated` can have either value. So the value of `is_paginated` does not matter to the overall conditional, and you could remove it from both the first and third clauses, and then remove either copy of the clause too.

PR #148, opened 2018-04-16
Added linting instructions to README

sgnn7:
Added instructions for linting files in the repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,10 +13,25 @@ LGPL v2.1
 
 ### Running tests
 - Prerequisites:
- - `npm install`
+ ```
+ $ npm install
+ ```
 
 - Running tests:
- - `npm test`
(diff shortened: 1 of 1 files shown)


manuq:
@sgnn7 two questions:

- Are we asking contributors to pass both checks, jshint and eslint?
- Besides checking a single file, is there a shortcut to make sure all the source code passes?

sgnn7:
@manuq I think we need to figure out which to use and run with it but as of now, both have configurations in the codebase. The shortcut to check all files would probably be `eslint .` in `lib/` or `jshint .` in root of project.

manuq:
@sgnn7 sounds good, let's merge this for now, but deprecate one of them to avoid one more requirement for potential contributors.

sgnn7:
@manuq Sounds good

PR #149, opened 2018-04-18
T21993

manuq:
https://phabricator.endlessm.com/T21993

### DIFF
--- a/examples/ars.js
+++ b/examples/ars.js
@@ -3,9 +3,6 @@
 // Ars Technica, Open Source tag example
 
 const Libingester = require('libingester');
-const thenifyAll = require('thenify-all');
-const RssToJson = thenifyAll(require('rss-to-json'), {}, ['load']);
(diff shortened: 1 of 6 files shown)


manuq:
@sgnn7 we started revisioning the `package-lock.json` when we automated the releases. I think is good to keep it, this way we ensure all developers and users have the exact same dependency tree. From [npm docs](https://docs.npmjs.com/files/package-lock.json): "This file is intended to be committed into source repositories".

sgnn7:
@manuq Eh - I don't quite agree with their recommendation there tbqh since what you gain in reproducibility you end up losing in API-compatible improvements, fixes, and security patches to your libraries. If the npm modules are abiding by the rules of versioning, it should be safe to let npm get you the best api-compatible version and I'd lean towards that side of the fence but I'm not primarily maintaining this codebase so it's not really my call. Maybe we can do a quick chat abotu this on slack at some point?

PR #152, opened 2018-04-18
Updated the linting rules from the ones in hatch-previewer

sgnn7:
This will bring them both in sync now


sgnn7:
@manuq Majority of the other PRs are based on the earlier rule that we already merged since I think this one will take a bit to get a consensus on

sgnn7:
@manuq Will do - I'll rebase this PR and add a commit that does something like that.

ptomato:
You can use [`difflint`](https://github.com/endlessm/difflint) to avoid adding the ignore comments and only check the code that was changed in the PR, but point taken.

PR #155, opened 2018-04-19
Linting and structure improvements to lib/verify.js

sgnn7:
With new linting rules we can now improve the verification module.
Additional fixes to structure and standards were done as well along with
sturdier assert logic.

### DIFF
--- a/lib/verify.js
+++ b/lib/verify.js
@@ -1,82 +1,101 @@
 'use strict';
 
+const fileType = require('file-type');
 const Validator = require('validator');
-const FileType = require('file-type');
-const {logger} = require('./logging');
+
+const { logger } = require('./logging');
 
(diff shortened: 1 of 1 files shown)


lib/verify.js
```diff
@@ -1,82 +1,101 @@
...
         }
     }
 }
 
-exports.VerificationError = VerificationError;
-
-function assert(value, message, metadata=null) {
-    if (value) {
+function assert(value, message, metadata = null) {
+    if (value === true) {
```
manuq: Do we have a lint rule that requires explicit check for true/false instead of using truthy/falsy ? Or is it needed for this validation only?

sgnn7: This check made no sense to be truthy since we always expect booleans to be passed in so I put in a `true` equality. It wasn't a linting issue.

PR #156, opened 2018-04-19
Improved linting and structure of lib/util.js

sgnn7:
Fixed majority of linting erros, converted internal representation to
camelCase, fixed inefficiencies, and updated tests to match.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -5,18 +5,35 @@
 const cheerio = require('cheerio');
 const crypto = require('crypto');
 const feedparser = require('feedparser');
-const libingester = require('./index');
 const parseDataUrl = require('parse-data-url');
 const request = require('request');
-const somaDOM = require('./soma_dom');
 const url = require('url');
-const {logger} = require('./logging');
+
 const config = require('./config');
+const libingester = require('./index');
+const { logger } = require('./logging');
+const somaDOM = require('./soma_dom');
 
 const USER_AGENT = 'libingester';
 
-const YOUTUBE_EMBED_DOMAINS = ['youtube.com', 'www.youtube.com',
-                               'www.youtube-nocookie.com'];
+const YOUTUBE_EMBED_DOMAINS = [ 'youtube.com',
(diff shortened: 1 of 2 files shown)


lib/util.js
```diff
@@ -5,18 +5,35 @@
...
 
 const USER_AGENT = 'libingester';
 
-const YOUTUBE_EMBED_DOMAINS = ['youtube.com', 'www.youtube.com',
-                               'www.youtube-nocookie.com'];
+const YOUTUBE_EMBED_DOMAINS = [ 'youtube.com',
+                                'www.youtube.com',
+                                'www.youtube-nocookie.com' ];
+
+const SAFE_ENCODING_CHARS = (
```
manuq: +1 to move this constant outside the function.

lib/util.js
```diff
@@ -31,35 +48,36 @@ let pooledRequest = request.defaults(requestDefaults);
 
 
 // Wrap request results in a promise
-function promisifiedRequest (options) {
+function promisifiedRequest(options) {
     return new Promise((resolve, reject) => {
-        pooledRequest(options, (err, res, body) => {
+        pooledRequest(options, (err, res) => {
```
manuq: Good, the third argument `body` is not used.

lib/util.js
```diff
@@ -435,104 +450,116 @@ function _fetch_rss_json_retrying(feed_url, attempt=1) {
...
     if (Array.isArray(feed)) {
-        return feed.map(uri => _do_create_wordpress(uri));
+        return feed.map(uri => _doCreateWordpress(uri));
     }
-    return _do_create_wordpress(feed);
+    return _doCreateWordpress(feed);
 }
 
-exports.create_wordpress_paginator = create_wordpress_paginator;
+module.exports = {
```
manuq: I understand is a bug to mix `module.exports` with `exports`. But what is the advantage of using `module.exports` over `exports`, as we've been doing? Just curious.

PR #158, opened 2018-04-20
Various improvements to lib/index.js

sgnn7:
Linting and structure improvements all around and general cleanup of
obvious code smells.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,22 +1,22 @@
 'use strict';
 
+const crypto = require('crypto');
+const path = require('path');
+
 const aws = require('aws-sdk');
 const cheerio = require('cheerio');
-const crypto = require('crypto');
+const fs = require('fs-extra');
 const mustache = require('mustache');
-const path = require('path');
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -1,23 +1,25 @@
...
+const tar = require('tar');
+
 const config = require('./config');
 const { logger } = require('./logging');
+const verify = require('./verify');
 
-const { promisify } = require('util');
-const fs = require('fs-extra');
-['readFile', 'writeFile', 'readdir', 'copy'].forEach((fn) => {
+// XXX: Aren't these already promisified with fs-extra?
```
manuq: You are right. I was so dumb to try `util.promisify` with standard `fs`. Please remove this wrapper and the `const { promisify } = require('util');` above.

sgnn7:
@manuq Tests still needed writeFile to be promisified for some reason so left it that way but I removed the others and the loop. We can figure out what the problem with it later so I'll merge it with only those changes.

PR #159, opened 2018-04-23
Refactor: Module split-out

sgnn7:
Moved classes into their own files and removed cyclic dependency. Also moved out data files into their own directory and cleaned up even more linting errors.

Part of (but not directly the implementation of)
https://phabricator.endlessm.com/T21994


lib/asset/baseAsset.js
```diff
@@ -0,0 +1,296 @@
...
+            const $asset = cheerio(this);
+
+            const assetId = $asset.attr('data-libingester-asset-id');
+            if (assetId) {
+                $asset.attr('data-soma-job-id', assetId);
+                $asset.attr('data-libingester-asset-id', null);
+            }
+
+            if (!$asset.attr('data-soma-job-id')) {
+                throw new ValidationError('Document has media without associated asset. ' +
```
manuq: I don't see why this validation is not part of the validation.js module. But we can change that later.

manuq: Nothing, I confused `ValidationError` with `VerificationError`. They look similar.

PR #162, opened 2018-04-25
Fix names

manuq:
- Move exported names back to original name. This fixes the documentation. Note: I haven't reverted the attribute names, assuming no user has been calling the API with the destructure assignment. Example: `get_embedded_video_asset({videoTag, videoUri})` VS `get_embedded_video_asset({video_tag, video_uri})`.
- Small style changes, and flex eslint rules. This PR makes eslint pass with zero errors or warnings.


lib/asset/baseAsset.js
```diff
@@ -61,9 +61,7 @@ class BaseAsset {
 
 
     /**
-     * 20-byte hex ID of the document
-     * @type string
-     * @readonly
+     * @returns {string} - 20-byte hex ID of the document
```
ptomato: Are you sure about this? `asset_id` is a property, so it seems to me that the jsdoc rule should not warn about this.

lib/config.js
```diff
@@ -63,7 +63,7 @@ const _config = {};
  *  - '--config-file' - A JSON config file with options inside. See above.
  * @memberof config
  */
-function parseOptions(options = {}) {
+function parse_options(options = {}) {
```
ptomato: My personal preference is to format JS default parameters like Python default parameters, e.g. `parse_options(options={})` No big deal either way.

lib/asset/baseAsset.js
manuq: You are right. Eslint was complaining but I will check why.

lib/config.js
manuq: This is the `space-infix-ops` Eslint rule. Gives: `Infix operators must be spaced`. I will change it.

lib/asset/baseAsset.js
manuq: The exact Eslint error was js-doc syntax error. It was because eslint wants `@type {string}`, not `@type string`. I reverted to previous state and added the curly brackets.

manuq:
@ptomato this is ready for review again.

lib/config.js
manuq: Yeah, the `space-infix-ops` rule doesn't have a more granular configuration. Is either on or off. I will turn it back on reverting the last commit.

manuq:
Following this in https://phabricator.endlessm.com/T22264

sgnn7:
Since all of @ptomato comments were fixed or mitigated and it lgtm, I'll merge this. CR pass!

PR #163, opened 2018-04-26
config: Change behaviour of settings passed to Hatch

manuq:
Merge the options passed to the Hatch constructor with the options
passed in the command line.

The original behaviour was to fully override the command line options
with the options passed to the Hatch constructor.

Added tests for command line options.  The config documentation is now
a bit easier to read.

https://phabricator.endlessm.com/T22232

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -44,13 +44,10 @@ const _config = {};
  *
  *   1. options.argv passed as parameter.
  *
- *   2. arguments passed to the command line. If 1. is passed, these
- *      arguments get fully overriden.
+ *   2. Arguments passed to the command line.
  *
- *   3. a JSON config file with options inside. Must be passed in 2.
- *      or 1. as `--config-file`. If more options are passed, they
- *      each override the same option if it is declared in the config
- *      file.
+ *   3. Options inside a JSON config file. Must be passed in 2. or
+ *      1. as `--config-file FILENAME`.
  *
  * @param {Object} options [{}] - A dictionary with options
  * @param {Array} options.argv - An array of arguments. Defaults to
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -37,20 +37,29 @@ const DEFAULT_OPTIONS = {
 const _config = {};
 
 
+function mergeOptions(optionsA, optionsB) {
```
sgnn7: Up to you but maybe this can be simplified a bit for clarity if argv can work with empty input (which I would assume it does). I also made a slight change since usually you are applying a change from one dictionary onto the other in the names of vars:
```
function mergeOptions(defaultOptions, optionsToMerge) {
    return Object.assign(argv(CLI_OPTIONS, defaultOptions),
                                     argv(CLI_OPTIONS, optionsToMerge));
}
```

test/lib/config.test.js
```diff
@@ -44,4 +48,19 @@ describe('config', () => {
...
+    it('can pass command line arguments', () => {
+        process.argv = [ 'node', 'my-ingester', '--path', '/tmp/mypath' ];
+        config.parse_options();
+        expect(config.get_setting('path')).to.equal('/tmp/mypath');
+    });
+
+    it('updates command line arguments', () => {
+        process.argv = [ 'node', 'my-ingester', '--path', '/tmp/mypath',
+                         '--no-tgz' ];
+        const options = { argv: [ '--path', './myNewPath' ] };
```
sgnn7: I still feel that ingesters shoudl add options directly to the options object and not `argv` but it's too big of a change right now so not a blocker:
```
options = {
  path: '/foo/bar',
}
```

PR #169, opened 2018-05-04
libingester: Encyclopedia

mazen-asef:
Add Encyclopedia object with more relevant parameters.
Copy NewsArticles templates for the encyclopedia ones
and modify index requirement list with the changes.

https://phabricator.endlessm.com/T21550

### DIFF
--- a/data/assets/encyclopedia-article.mst
+++ b/data/assets/encyclopedia-article.mst
@@ -1,0 +1,62 @@
+<!doctype html>
+<html>
+<head>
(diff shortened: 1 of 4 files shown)


ptomato:
I think the EncyclopediaArticle should be part of the Encyclopedia ingester, rather than libingester.

manuq:
I agree with @ptomato, the EncyclopediaArticle asset, template and stylesheet can all be part of the ingester codebase.

PR #171, opened 2018-05-08
Allow setting multiple fields at once

manuq:
https://phabricator.endlessm.com/T22483

### DIFF
--- a/lib/asset/baseAsset.js
+++ b/lib/asset/baseAsset.js
@@ -212,7 +212,7 @@ class BaseAsset {
     /**
      * @returns {Object}
      */
-    to_metadata() {
+    to_hatch_metadata() {
         // Build an object with all required metadata
         const metadata = {
             'assetID': this._asset_id,
@@ -292,7 +292,7 @@ class BaseAsset {
         hatch._assets.push(this);
         hatch._promises.push(this._process().then(() => {
             const data = this.to_data();
-            const metadata = this.to_metadata();
+            const metadata = this.to_hatch_metadata();
 
             if (data) {
                 metadata.cdnFilename = `${metadata.assetID}.data`;
--- a/lib/asset/baseAsset.js
+++ b/lib/asset/baseAsset.js
@@ -18,12 +18,15 @@ class BaseAsset {
         return 'EknStaticTag';
     }
 
-    constructor(verifier = assetVerifier) {
+    constructor(metadata = {}, verifier = assetVerifier) {
         this._asset_id = this._newAssetId();
         this._ekn_tags = [];
+        this._verifier = verifier;
         this._tags = [];
 
-        this._verifier = verifier;
+        if (metadata) {
+            this.set_metadata(metadata);
+        }
     }
 
     _newAssetId() {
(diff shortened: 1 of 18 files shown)


lib/asset/baseAsset.js
```diff
@@ -69,6 +72,41 @@ class BaseAsset {
...
+        if (fields.canonical_uri) {
+            this.set_canonical_uri(fields.canonical_uri);
+        }
+        if (fields.last_modified_date) {
+            this.set_last_modified_date(fields.last_modified_date);
+        }
+        if (fields.date_published) {
+            this.set_date_published(fields.date_published);
+        }
+        if (fields.sequence_number) {
```
ptomato: These should probably distinguish between "field not present" and "field set to falsey value". It might not matter that much for the strings, but `sequence_number` can legitimately be 0.

lib/asset/blogArticle.js
```diff
@@ -13,11 +13,42 @@ const { ensureCheerio, fixLinkTargets } = require('./util');
...
+        if (fields.body) {
+            this.set_body(fields.body);
+        }
+        if (fields.custom_scss) {
+            this.set_custom_scss(fields.custom_scss);
+        }
+        if (fields.read_more_text) {
+            this.set_read_more_text(fields.read_more_text);
+        }
+        if (fields.as_static_page === true) {
```
ptomato: Here the opposite comment, you can accept any truthy value.

lib/asset/blogArticle.js
```diff
@@ -13,11 +13,42 @@ const { ensureCheerio, fixLinkTargets } = require('./util');
 
 /** */
 class BlogArticle extends BaseAsset {
-    constructor() {
-        super();
+    constructor(fields) {
```
ptomato: I'd suggest adding default parameters to all the constructors: `fields={}`

lib/asset/blogArticle.js
```diff
@@ -13,11 +13,42 @@ const { ensureCheerio, fixLinkTargets } = require('./util');
...
+        super(fields);
+
         this._object_type = 'ArticleObject';
         this._content_type = 'text/html';
-        this._tags = [];
+    }
+
+    /**
+     * Set multiple fields at once.
+     * @param {Object} fields - Contains See each setter below for a
```
ptomato: Typo: remove `Contains`

ptomato: Optionally, it might give better documentation to do something like
```
 * @param {string} fields.author - See @link{set_author}
 * @param {string} fields.title - See @link{BaseAsset.set_title}
```
etc.

lib/asset/baseAsset.js
manuq: Good catch!

lib/asset/blogArticle.js
manuq: Thanks for noticing.

manuq:
Regarding the use of "fields", I started using metadata, but then I found the `to_metadata` current method, which converts all these into the Hash metadata. I don't have a problem moving all back to metadata, but I think we should make a distinction.

manuq:
@ptomato I see the Asset `to_metadata()` is not API, is used from the Hash. Maybe we can rename it `to_hash_metadata`?

manuq:
I made the requested changes. The missing bit is what to do with the "fields" wording. Should we rename to metadata?

ptomato:
I'm fine with renaming `to_metadata()` to something else.

manuq:
Good. I updated the commits and appended a new one with the rename.

lib/asset/baseAsset.js
```diff
@@ -18,12 +18,15 @@ class BaseAsset {
...
 
-    constructor(verifier = assetVerifier) {
+    constructor(metadata = {}, verifier = assetVerifier) {
         this._asset_id = this._newAssetId();
         this._ekn_tags = [];
+        this._verifier = verifier;
         this._tags = [];
 
-        this._verifier = verifier;
+        if (metadata) {
```
cosimoc: Are you sure `{}` evaluates to false-y? Since you have a default value for the parameter, is this code path not always taken? Maybe I'm missing something

lib/asset/baseAsset.js
```diff
@@ -69,6 +72,51 @@ class BaseAsset {
...
+        }
+        if (typeof metadata.sequence_number !== 'undefined') {
+            this.set_sequence_number(metadata.sequence_number);
+        }
+        if (typeof metadata.license !== 'undefined') {
+            this.set_license(metadata.license);
+        }
+        if (typeof metadata.tags !== 'undefined') {
+            this.set_tags(metadata.tags);
+        }
```
cosimoc: I think it would be nice to print out warnings, or even error out, for any metadata key that was passed-through but ends up not being used.

manuq: +1

lib/asset/baseAsset.js
manuq: You are right, this is always going through.

lib/asset/baseAsset.js
```diff
@@ -18,12 +18,20 @@ class BaseAsset {
         return 'EknStaticTag';
     }
 
-    constructor(verifier = assetVerifier) {
+    get ALLOWED_METADATA() {
```
ptomato: could be `static get`?

lib/asset/blogArticle.js
```diff
@@ -13,11 +13,57 @@ const { ensureCheerio, fixLinkTargets } = require('./util');
 
 /** */
 class BlogArticle extends BaseAsset {
-    constructor() {
-        super();
+    get ALLOWED_METADATA() {
```
ptomato: Same here, `static get`? (and `BaseAsset.ALLOWED_METADATA` below)

lib/asset/baseAsset.js
manuq: +1

lib/asset/blogArticle.js
manuq: I will move to `static get`. Please note, after running `eslint` I moved the `if {} else if {}` to `switch/case` statements.

manuq: @ptomato now I remember why I didn't use `static get` for the allowed metadata. The children assets expand the metadata allowed. So the `_check_metadata_is_allowed` uses `this.ALLOWED_METADATA` which is different depending on the asset. If I use `static get` I will have to reimplement `_check_metadata_is_allowed` in each asset.

PR #173, opened 2018-05-15
util: Add new utils fetch_news_article(), fetch_blog_article()

manuq:
To handle HTTP response errors and mark the Article assets
corresponding to the failing URL as failed.  This way the assets are
accounted in the % of failed assets.  Typical usage is:

    function ingest_article(hatch, uri) {
        return libingester.util.fetch_news_article(hatch, uri).then(({ $, asset }) => {
            if (asset.has_failed()) {
                return;
            }
            // normal article ingestion
        });
    }

https://phabricator.endlessm.com/T22195

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -14,7 +14,9 @@ const imageType = require('image-type');
 const config = require('./config');
 const { logger } = require('./logging');
 const somaDOM = require('./soma_dom');
+const BlogArticle = require('./asset/blogArticle');
 const ImageAsset = require('./asset/imageAsset');
+const NewsArticle = require('./asset/newsArticle');
 const VideoAsset = require('./asset/videoAsset');
 
 const USER_AGENT = 'libingester';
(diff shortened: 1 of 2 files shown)


lib/asset/baseAsset.js
```diff
@@ -242,6 +243,11 @@ class BaseAsset {
         }
 
         hatch._assets.push(this);
+        if (this._has_failed) {
+            hatch._failed_assets.push(this);
```
cosimoc: Is it OK that we are putting this asset both in the regular and the failed lists?

lib/asset/baseAsset.js
```diff
@@ -296,6 +302,21 @@ class BaseAsset {
         }
         return assets;
     }
+
+    /**
+     * Explicitely mark this asset as failed
```
cosimoc: Typo: `Explicitly`

lib/asset/baseAsset.js
```diff
@@ -296,6 +302,21 @@ class BaseAsset {
...
+     */
+    set_has_failed() {
+        this._has_failed = true;
+    }
+
+    /**
+     * Check if this asset has been explicitely marked as failed
```
cosimoc: Ditto

lib/util.js
```diff
@@ -16,6 +16,8 @@ const { logger } = require('./logging');
 const somaDOM = require('./soma_dom');
 const ImageAsset = require('./asset/imageAsset');
 const VideoAsset = require('./asset/videoAsset');
+const NewsArticle = require('./asset/newsArticle');
+const BlogArticle = require('./asset/blogArticle');
```
cosimoc: Please keep these in alphabetical order

lib/util.js
```diff
@@ -573,6 +628,8 @@ module.exports = {
     download_youtube_thumbnail,
     encode_uri,
     fetch_html,
+    fetch_news_article,
+    fetch_blog_article,
```
cosimoc: Alphabetical order

lib/util.js
```diff
@@ -151,6 +153,59 @@ function fetch_html(uri, encoding) {
...
 
+async function _fetch_article(hatch, uri, encoding, Article) {
+    const asset = new Article();
+    let $;
+    try {
+        $ = await fetch_html(uri, encoding);
+    } catch (error) {
+        if (error.message.includes('non-successful status code')) {
+            logger.error(error);
+            asset.set_has_failed();
```
cosimoc: Instead of setting this flag here on the asset, couldn't we directly add the asset to the failed list in the hatch? We could remove the awkward `set_has_failed()` API

lib/asset/baseAsset.js
```diff
@@ -296,6 +302,21 @@ class BaseAsset {
...
+     * @returns {Boolean} - true/false
+     */
+    has_failed() {
```
cosimoc: This does not seem to be used

lib/util.js
manuq: Yes this `set_has_failed()` API is awkward. It is just for checking in the ingester:
```
        if (asset.has_failed()) {
            return;
        }
```
But I can add a new API for the Hatch, `Hatch.has_failed_asset(asset)` instead.

lib/asset/baseAsset.js
manuq: Yes, the successful assets are calculated by filtering failed from regular, and then the % of failed is calculated as successful / regular .

lib/util.js
manuq: Will do. This is not being checked in our lint rules. Do you think we should enable https://eslint.org/docs/rules/sort-imports ?

lib/util.js
manuq: This is also not being checked in our lint rules. Do you think we should enable sort-keys? https://eslint.org/docs/rules/sort-keys

lib/util.js
cosimoc: Yeah, I would like to. However I'd move linting cleanups to a different PR.

lib/util.js
cosimoc: Yes, but see above.

PR #174, opened 2018-06-08
Mobile rendering: Set font sizes and margin for mobile

zarkooroz:
https://phabricator.endlessm.com/T22646


data/assets/gallery-image-stylesheet.scss
```diff
@@ -139,3 +139,20 @@ h1 {
         font-family: $support-font-composite;
     }
 }
+
+/* Smartphones (portrait and landscape) */
+@media only screen 
+and (min-device-width: 320px) 
+and (max-device-width: 719px) {
+    html {
+        font-size:16px;
```
erikos: Shouldn't we use the mixin for this?

@include responsive-elements($body-size: 16px);

erikos:
Updated the branch to use the mixin.

manuq:
@erikos only one thing needed: override font size for the h2-h6 elements just like the News stylesheet.

PR #177, opened 2018-06-15
util: Add utility function to clean articles body

manuq:
https://phabricator.endlessm.com/T22968

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -620,7 +620,76 @@ function create_wordpress_paginator(feed) {
     return _doCreateWordpress(feed);
 }
 
+/**
+ * Default options for {@link utilcleanup_body}
+ *
+ * @property {Array} remove
+ * @property {Array} removeNoText
+ * @property {Object.<string, Array>} removeAttrs
+ * @property {boolean} noComments
+ * @memberof util
+ */
+const CLEANUP_DEFAULTS = {
+    remove: [ 'script', 'noscript', 'style', 'iframe', 'hr', 'br + br', 'center' ],
+    removeNoText: ['p'],
+    removeAttrs: { '*': [ 'style', 'class' ] },
+    noComments: true,
+};
+
+/**
+ * Utility to remove the article body.
+ *
+ * The cleanup should be among the last operations done with the
+ * body. For example, if your ingester tries to make Video assets from
+ * iframes, the cleanup utility removes iframe elements by default.
+ *
+ * @param {Cheerio} body - The Cheerio object that will be cleaned
+ *   up. Is called body because it usually represents an article body.
+ * @param {Object} options - Options to configure the cleanup, by
+ *   default {@link utilCLEANUP_DEFAULTS} is used. You can expand or
+ *   override the defaults.
+ * @param {Array} options.remove - Array of CSS selectors. The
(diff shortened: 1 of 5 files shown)


lib/util.js
```diff
@@ -620,7 +620,78 @@ function create_wordpress_paginator(feed) {
...
+ * @property {Array} remove - The default values for ....
+ * @property {Array} removeNoText - The default values for ...
+ * @property {Array} removeAttrs - The default values for ...
+ * @property {boolean} noComments - The default values for ...
+ * @memberof util
+ */
+const CLEANUP_DEFAULTS = {
+    remove: [ 'script', 'noscript', 'style', 'iframe', 'hr', 'br + br', 'center' ],
+    removeNoText: ['p'],
+    removeAttrs: [[ ['*'], [ 'style', 'class' ] ]],
```
ptomato: I would suggest making the removeAttrs api to take a dictionary, e,g.:
```js
{'*': ['style', 'class']}
```
I know that would disallow things that are possible with the current API, like `[[['b', 'i']: ['role']]]` but the dictionary style seems to me a lot easier to think about. Arrays of arrays of arrays are difficult to grasp in my experience.
Otherwise, maybe something like this:
```js
[
  { tags: ['b', 'i'], attrs: ['class', 'style'] },
]
```

lib/util.js
```diff
@@ -620,7 +620,78 @@ function create_wordpress_paginator(feed) {
...
+ * @param {Array} options.removeAttrs - Use this to remove attributes
+ *   from elements. ‘removeAttrs’ is an Array of pairs [ tags, attrs
+ *   ], where ’tags’ is an Array of CSS selectors, and ‘attrs’ is an
+ *   Array of attributes.
+ * @param {boolean} options.noComments - If true (default), all HTML
+ *   comments will be removed from the body.
+ * @memberof util
+ */
+function cleanup_body(body, {
+    remove = CLEANUP_DEFAULTS.remove,
```
ptomato: Nice use of defaults!

lib/util.js
manuq: Good point, I will go with the dictionary API. It doesn't disallow multiple tags, because the passed CSS selector can be: `{'b, i': ['style', 'class']}`

manuq:
I changed the API to use a dictionary for removing attributes, and also added unit tests.

lib/util.js
```diff
@@ -620,7 +620,76 @@ function create_wordpress_paginator(feed) {
...
+ * Utility to remove the article body.
+ *
+ * The cleanup should be among the last operations done with the
+ * body. For example, if your ingester tries to make Video assets from
+ * iframes, the cleanup utility removes iframe elements by default.
+ *
+ * @param {Cheerio} body - The Cheerio object that will be cleaned
+ *   up. Is called body because it usually represents an article body.
+ * @param {Object} options - Options to configure the cleanup, by
+ *   default {@link utilCLEANUP_DEFAULTS} is used. You can expand or
```
dnohales: Missing dot here?

manuq: No, that's on purpose. The link in the documentation doesn't work otherwise.

dnohales: Node...

PR #183, opened 2018-06-27
util: Add utility to extend cleanup defaults

manuq:


lib/util.js
```diff
@@ -662,10 +662,70 @@ const CLEANUP_DEFAULTS = {
...
+    removeData = [],
+    removeAttrs = {},
+    noComments,
+} = {}) {
+    const mergedRemoveAttrs = _mergeObjectArrays(CLEANUP_DEFAULTS.removeAttrs, removeAttrs);
+
+    let mergedNoComments;
+    if (typeof noComments === 'undefined') {
+        mergedNoComments = CLEANUP_DEFAULTS.noComments;
+    } else {
```
dnohales: I'd sound like Srdjan but you can remove the `else` by just doing this above:

```
let mergedNoComments = noComments;
```

lib/util.js
```diff
@@ -662,10 +662,70 @@ const CLEANUP_DEFAULTS = {
...
+/**
+ * Utility to extend {@link utilcleanup_body} defaults.
+ *
+ * @param {Object} options - Pass an Object with the same format as
+ *   {@link utilCLEANUP_DEFAULTS}.
+ * @returns {Object} - This is a merge of the {@link
+ *   utilCLEANUP_DEFAULTS} and the passed options.
+ * @memberof util
+ */
+function extend_cleanup_defaults({
```
dnohales: What about `extend_cleanup_body_defaults` to keep consistency.

manuq: Hmm I tried and started to port ingesters, but is too long. The variable with defautls is called `CLEANUP_DEFAULTS` so I think prepending `extend_...` to it is OK.

PR #190, opened 2018-07-24
Don't save the title in the manifest

dnohales:
This is going to break compatibility with the current hatch structure.
But it's really needed to save a considerable amount of memory on big
ingesters (encyclopedia).

FTR, the title from the manifest file is not actually used in SOMA.


manuq:
+1 I confirm this change doesn't affect HatchPreviewer. or basin-hatch . Merging.

PR #197, opened 2018-09-11
T23760 tests

manuq:
https://phabricator.endlessm.com/T23760


lib/utils/cleanup-html.js
```diff
@@ -42,7 +42,7 @@ function _removeMatchingAttributes ($html, selector, regex) {
 
 function cleanupHtml ($html, {
     remove = cleanupHtmlDefaults.remove,
-    removeNoText = cleanupHtmlDefaults.removeNoText,
+    // removeNoText = cleanupHtmlDefaults.removeNoText,
```
dnohales: Commited by accident?

package.json
```diff
@@ -35,21 +33,14 @@
...
-    "rewire": "^3.0.2",
-    "sinon": "^4.4.10"
+    "jest": "^23.6.0"
   },
   "scripts": {
     "docs:build": "rm -rf docs && documentation build lib/index.js lib/util.js lib/config.js -o docs -f html",
     "docs:publish": "npm run docs:build && git add -A docs && git commit -m 'Auto commit documentation'",
     "release": "./run_release.sh",
-    "test": "./run_tests.sh"
+    "test": "npx jest"
```
dnohales: You don't need to use `npx` within `package.json`

test/lib/processors/normalize-headers-processor.test.js
```diff
@@ -0,0 +1,30 @@
...
+    const out = await readHtml('normalize-headers-processor.out-1.html');
+    const result = libingester.processors.processNormalizeHeaders($body);
+    expect(result.$body.html()).toBe(out);
+});
+
+test('does not pass h6', async () => {
+    const out = await readHtml('normalize-headers-processor.out-2.html');
+    const result = libingester.processors.processNormalizeHeaders($body, { fromIndex: 5 });
+    expect(result.$body.html()).toBe(out);
+});
```
dnohales: What you normally do is to use [describe](https://jestjs.io/docs/en/api#describename-fn) to group tests if they are related, they are going to be grouped afterwards in the output.

lib/utils/cleanup-html.js
manuq: No. It is in the lint commit to make the lint pass, because the `removeNoText` below is also commented. I plan to remove this anyway, is coming from the v2 cleanup but not very useful because it can remove other elements inside.

PR #198, opened 2018-09-13
T23760 more tests

manuq:
https://phabricator.endlessm.com/T23760


examples/prensa-libre.js
```diff
@@ -48,18 +48,20 @@ class PrensaLibreParser extends libingester.HTMLArticleParser {
...
+                     '[data-desktop*=".advice-wrap"]',
+                     '[data-mobile*=".floating-advice"]',
+                     '[data-tablet*=".floating-advice"]',
+                     '.subscribe-module',
+                     '#divInline_Notas',
+                     'article',
+                     'br',
+                 ],
+             },
+             }],
```
dnohales: Let's reindent this.

lib/html-article-parser.js
```diff
@@ -125,16 +134,18 @@ class HTMLArticleParser extends Parser {
         return [
             processImages,
             processYoutubeEmbeds,
-            processNormalizeHeaders,
-            processCleanup({}, true),
+            normalizeHeaders,
+            [ canonicalizeLinks, { baseUrl: this.parseCanonicalURI() } ],
```
dnohales: We are yet to define whether we have to canonicalize links in future pipeline, I disagree on doing it. Plus, this is not canonicalizing the URIs since only the linked article knows its canonical URI, this is a mere conversion from relative to absolute, which doesn't add anything useful down to the SDK.

I'd prefer to keep it on hold until we reach consensus.

lib/html-article-parser.js
```diff
@@ -125,16 +134,18 @@ class HTMLArticleParser extends Parser {
...
+            [ processCleanup, { useDefaults: true } ],
         ];
     }
 
     async processBody ($body) {
         let $processedBody = $body;
 
         for (const processor of this.bodyProcessors) {
-            const result = await processor.bind(this)($processedBody);
+            const proc = this._applyOptions(processor);
```
dnohales: What about `processorFunction`?

lib/html-article-parser.js
```diff
@@ -30,6 +31,14 @@ class HTMLArticleParser extends Parser {
         return encodeURI(url.format(url.parse(uri)));
     }
 
+    _applyOptions (processor) {
+        if (_.isArray(processor)) {
+            const [ proc, options ] = processor;
```
dnohales: What about `processorFunction`?

examples/prensa-libre.js
manuq: To 4 spaces? OK, current lint rules don't complain about this.

lib/html-article-parser.js
manuq: Fair. I think this is useful in any case.

lib/html-article-parser.js
manuq: +1

lib/html-article-parser.js
manuq: +1

examples/prensa-libre.js
dnohales: I mean, it's not properly indented, look at it :)

PR #199, opened 2018-09-18
Ingester: log total / failed, and exit accordingly

manuq:
RC will be 0 for none assets failed, 1 for many assets failed, 13 for
some assets failed. The failure rate threshold is configurable, and
defaults to 90% of the assets.

By default the ingester run() will exit the process. Passing exit =
false you can skip this.

https://phabricator.endlessm.com/T23489

### DIFF
--- a/lib/ingester.js
+++ b/lib/ingester.js
@@ -9,6 +9,7 @@ const { ImplementationError } = require('./errors');
 class Ingester {
     constructor (ingesterPath = null) {
         this.ingesterPath = ingesterPath;
+        this.assetsCount = { 'total': 0, 'failed': 0 };
 
         if (!this.ingesterPath) {
             this._name = 'libingester';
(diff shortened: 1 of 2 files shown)


lib/ingester.js
```diff
@@ -36,27 +37,46 @@ class Ingester {
...
 
+    getIngestionStatus () {
+        const failedProportion = this.assetsCount.failed / this.assetsCount.total;
+        if (failedProportion === 0) {
+            return { 'status': 'OK', 'rc': 0 };
+        }
+        if (failedProportion > config.hatchFailRateThreshold) {
+            return { 'status': 'FAILED', 'rc': 1 };
+        }
+        return { 'status': 'WARNING', 'rc': 13 };
```
dnohales: Out of curiosity, what `rc` means? maybe `exitCode` or `errno` are better names?

manuq: I'll rename rc to returnCode

PR #203, opened 2019-02-12
Make revisionTag optional and allow to override it

dnohales:
Sometimes the revisionTag can be another value rather than just the last
modified date (like the ETag or something like that), let's allow the
ingester to override the revisionTag for the asset.

Also, since the Soma pipeline is adding support to calculate the
revisionTag using hashing, let's allow the ingester to not specify
the revisionTag at all, so it's calculated down the pipeline.

### DIFF
--- a/test/lib/asset/blogArticle.test.js
+++ b/test/lib/asset/blogArticle.test.js
@@ -88,6 +88,20 @@ describe('BlogArticle', () => {
         const hatchMetadata = asset.to_hatch_metadata();
         expect(hatchMetadata.synopsis).to.equal('This is a line. This is the same line.');
     });
+
+    it('can override revisionTag', () => {
+        asset.set_revision_tag('TestTag');
+        asset.render();
+        const hatchMetadata = asset.to_hatch_metadata();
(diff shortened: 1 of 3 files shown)


test/lib/asset/blogArticle.test.js
```diff
@@ -88,6 +88,20 @@ describe('BlogArticle', () => {
...
+        const hatchMetadata = asset.to_hatch_metadata();
+        expect(hatchMetadata.revisionTag).to.equal('TestTag');
+    });
+
+    it('can delete revisionTag', () => {
+        asset.set_revision_tag(false);
+        asset.render();
+        const hatchMetadata = asset.to_hatch_metadata();
+        expect(hatchMetadata).to.not.have.property('revisionTag');
+    });
```
mazen-asef: Just wondering if setting these tests on the other assets (e.g. NewsArticle) would be good too.

dnohales: Well, I don't think it's necessary, since the `revisionTag` is part of `BaseAsset`.

